PR #1, opened 2019-01-03
fix setup.py to support use as a python module

inoryy:
With this change an end-user can use `ffai` as an external python module, e.g.

```
$ pip install git+https://github.com/njustesen/ffai
...
Successfully built ffai
Installing collected packages: ffai
Successfully installed ffai-0.0.1
$ python
>>> import ffai
>>> ...
```
The web server can now be started by running `python -m ffai.web.server` (from any directory).

You can also build a wheel (e.g. to put on `pypi`) by running `python setup.py sdist bdist_wheel`.

**NB!** I had to move the `data` folder to `ffai/data` to ensure it's distributed through `pip`.


njustesen:
Thank you for improving FFAI :)
What's the idea behind having two data folders? One in the root and one within /ffai/?
I'll pull and test your changes later today.

inoryy:
@njustesen it's not two `data` folders -  it's the original one that I *moved* to `ffai/data` to make sure it's distributed correctly. When an end-user install FFAI through pip he will see the following:

```
$ pip install git+https://github.com/inoryy/ffai
...
Successfully built ffai
Installing collected packages: ffai
Successfully installed ffai-0.0.1
$ ls ~/anaconda3/lib/python3.6/site-packages/ffai
ai  core  data  __init__.py  __pycache__  web
```

PR #3, opened 2019-01-10
Fix small issue with observation 'state' dictionary

FragLegs:
The observation state dictionary sets the key `own turn` twice, once to mean "is my team the currently active team?" and once to mean "how many turns has my team had?". The second version overwrites the first, unfortunately. This fix renames the "is my team the currently active team?" to `own active` so that both values can exist.


njustesen:
Thanks! I ended up renaming 'own turn' and 'opp turn' to 'own turns' and 'opp turns' instead.

PR #21, opened 2019-03-22
Better compatibility with FUMBBL

rreche:
In order to replay fumbbl games with ffai, I made the following modifications:

 - Invert the side of home team and away team
 - Home (resp. Away) team is the red (resp. blue) team
 - The scatter direction:
       Before:         After:
        123               812
        4X5               7X3 
        678               654


njustesen:
I am afraid it will break the UI because there are no changes to the web module. Have you checked if it works when playing in the web app? Also, do you have any updates on the replay module? I am very curious.

rreche:
Hi Niels,

It’s true, I have to test the web module.

My replay module is done but I have to test more games before submitting it. I will ask on Discord a list of match Orc vs Humans. I think I will need to add some code for the Ogre player.

I have also pushed another branch with some rule corrections (mainly tackle, interception and apothecary)

Rémy


Envoyé de mon iPhone

njustesen:
Any news on this? I would love to try the replay module.

PR #32, opened 2019-04-26
correct some  rules issues (Tackle and Dodge, Apothecary with KO)

rreche:
I found several minor errors which are corrected in this branch:

- a wrong condition to test is a play can catch or not
- a list of wrong conditions to test if an adjacent player with some skill exists or not
- when someone use an Apothecary on a KO, the player is still on the pitch but stunned (instead of being put on reserve)
- Tackle was not considered for dodge and block results

I have also done two minor changes and I hope it doesn't break anything:
- I change the order between WeatherTable and Fans procedure
- During Setup, I modify the setup reorganization

### DIFF
--- a/ffai/core/model.py
+++ b/ffai/core/model.py
@@ -448,19 +448,19 @@ class Pitch:
         diving_tacklers = []
         shadowers = []
         tentaclers = []
-        for square in self.adjacent_player_squares(player.position, include_own=False, include_opp=True):
+        for square in self.adjacent_player_squares(player, include_own=False, include_opp=True):
             player_at = self.get_player_at(square)
             if player_at is not None and player_at.has_tackle_zone():
                 tackle_zones += 1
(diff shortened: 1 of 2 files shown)


njustesen:
Very nice. I will test it soon. Btw. rreche, please get in touch on discord.

rreche:
I change my name on Discord from rr3my to rreche to be sure that people know that it is the same person.

Envoyé de mon iPhone

njustesen:
Can I try your replay module?

rreche:
I am on vacation until Sunday. I will sens it after that.

Envoyé de mon iPhone

njustesen:
Thanks.

njustesen:
Hi Rémy,
Sorry for being impatient, but will you be able to share the replay module sometime this week?
Thanks!

PR #71, opened 2019-11-08
Tests and minor fix for Mighty Blow

danj3000:
Added some tests for MB under injury and armour.
It turned out that mighty blow was being applied even when the roll doesn't require it, meaning it wasn't usable in armour rolls.
Fixed!

### DIFF
--- a/ffai/core/procedure.py
+++ b/ffai/core/procedure.py
@@ -148,10 +148,8 @@ class Armor(Procedure):
             # Armor broken
             if result >= roll.target:
                 armor_broken = True
-
-            # Armor broken - Might Blow
-            if self.inflictor is not None and self.inflictor.has_skill(Skill.MIGHTY_BLOW) \
-                    and result + 1 > self.player.get_av():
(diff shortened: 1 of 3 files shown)


DrRerNate:
Mighty Blow is a normal, non-mandatory skill, meaning the coach should be asked if it should be applied to the armor roll and then (provided armor was broken and Mighty Blow not used) asked again if it should be applied to the injury roll.

I'm not denying that 99.99% of the time you want to use the skill "the right way," but that should be a choice for the bot developers, not something that is forced upon them.

Besides, there is always that 0.01% case where you need the square to be filled with a player (knocked down or otherwise) in order to chain push through that square...

Same should apply to all other non-mandatory skills as well...

njustesen:
I think we should replicate what FUMBBL is doing. I know many skills are automatically applied. Not sure how it is in Cyanide. Ideally, we can make it an option in the configuration file.

danj3000:
I don't disagree. However, these tests catalog the existing behaviour, and fix to enable its use in armour rolls. So a better situation than before.

Issue #61, opened 2019-10-24
Prehensile tail to be added to dodge modifiers


danj3000:
This and Tackle (#55) need a list of players with tz's on the 'from' square to be passed into the dodge calculation

danj3000:
Fixed

PR #72, opened 2019-11-08
Adding prehensile tail to the dodge modifiers

danj3000:
turns out the hard work was done already, so just a case of wiring it through to the modifier calc.
Fixes issue #61 

There's a 6-part tuple returned by game.tackle_zones_in_detailed - could probably do with replacing with an explicit class.  It ran off down a rabbit hole into Pitch though, so I wasn't confident to refactor with abandon.

### DIFF
--- a/ffai/core/model.py
+++ b/ffai/core/model.py
@@ -588,28 +588,7 @@ class Pitch:
         return tackle_zones
 
     def get_tackle_zones_detailed(self, player):
-        tackle_zones = 0
-        tacklers = []
-        prehensile_tailers = []
-        diving_tacklers = []
-        shadowers = []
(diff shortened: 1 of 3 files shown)


danj3000:
I've added a class as suggested

danj3000:
I think the preferred usage is without the parameter. The only reason for the parameter is for later in Dodge when the player fails the dodge roll, to determine if any tacklers are preventing the skill use:
' 
               # Check if dodge
                if self.player.has_skill(Skill.DODGE) and (not self.dodge_used) and (not tacklers):
'
Probably this could be done better using a convenience method on game/pitch at the point of checking.

PR #74, opened 2019-11-15
Bonehead skill added

danj3000:
Adding the bonehead skill as a procedure - modelled on GFI
`start_player_action` in Turn creates the PlayerAction proc, and a Bonehead proc if the player has the 'skill'
Passing the PlayerAction object to the Bonehead proc allows Bonehead to set the 'done' flag if it triggers. EndPlayerTurn does the rest.
The game loop cleans up procs from the stack as normal.

This was the simplest way I could think of to implement the proc interactions without deviating from the existing code flow.


danj3000:
Just updated with a basic state icon for the front end
Also the bone_head state wasn't reset on a second successful test, so have added a test and fixed that.
Should be complete now.

PR #79, opened 2019-11-22
Adding wild animal and really stupid

danj3000:
added a couple of negatraits: WA and RS.

with very basic icons for UI


danj3000:
Ah, the classic lower case bon error!... I'll sort it.

Do you ever get more than one negatrait on a player? I don't want to go overboard, but I guess we would just trigger each in turn...?

njustesen:
The skills_used list could also be on the PlayerAction object, right? I am, however, not sure what is best. I'll merge this pull request after fix no. 1 above, then we can always improve later.

danj3000:
Fix #1 is good to go.
Skills used *could* be tracked on PlayerAction, but imo PlayerState seemed a natural home, and would support UI indicators if wanted (tenuous, but the first thing I could think of 🙂)

Issue #85, opened 2019-12-04
Decouple bots from the game engine stack


danj3000:
Procbots and grodbot are tightly coupled to the game procedure stack in order to determine what's going on in the game.

They do this by directly using game.state.stack.peek() to get the last procedure in the stack.

This kind of deep access is limiting how the game engine can be changed, and specifically prevents the introduction of procedures that don't directly reflect the current action of the game 

e.g. reroll or 'opponent input' procedures could both be created as part of an overall block game action, but do not themselves indicate that a block is taking place.

I suggest replacing access to the game stack with a method on Game that can walk the stack to determine the relevant procedure to return - ideally as an enum value. Any new generic procedure would be flagged so as to be skipped in the process.

njustesen:
I like the idea of having a method on Game instead of inspecting the stack. However, introducing enums that are 99% similar to the classes implementing Procedure seems redundant. 

What would this new function return e.g. during a Dodge procedure if Reroll is added to the stack? Should the Agent see that the current procedure is Dodge or Reroll?

danj3000:
I think for your example the bot needs to know the overall context - Dodge. The fact that it failed and is a reroll choice is implied by the available actions.

Regarding the use of classes or enums - I admit that it is more code for seemingly little benefit, but if I were designing an interface on Game from scratch I don't think I'd choose to return the actual objects on the stack rather than a string/enum/whatever to identify the domain meaning. This isn't 100% necessary at this stage though, so I'll put together a PR using the current proof approach.

njustesen:
Great. 

I agree that passing Procedure sub-classes is a weird way to inform about the stage of the game. The idea behind the ProcBot was to hide this complexity by overriding methods instead. In fact, the ProcBot should be the standard way of implementing a scripted bot in FFAI. Maybe it should be called something more user-friendly.

PR #86, opened 2019-12-06
Hiding the stack

danj3000:
This is the basic approach mentioned in issue #85 

I have replaced all calls to game.state.stack.peek() with a new method 'game.get_procedure_context()'. This returns the latest Procedure that dictates the game's context e.g. Dodge, Block etc... but ultimately should be changed to abstract away from the Procedure class details.

The Procedure class now has a field 'has_context' which is true by default, but for the new ReRoll procedure the value is set to False. This means that the ReRoll procedure is not dictating the game context, and should not be returned when get_procedure_context is called.

get_procedure_context simply walks down the stack looking for the first procedure where has_context is true.

new procedures: reroll and loner to demo how these would work as procedures - currently only triggered from the negatraits.

### DIFF
--- a/ffai/ai/bots/grodbot.py
+++ b/ffai/ai/bots/grodbot.py
@@ -104,6 +104,8 @@ class GrodBot(Agent):
             return self.dodge(game)
         if isinstance(proc, Pickup):
             return self.pickup(game)
+        if isinstance(proc, Negatrait):
+            return self.negatrait(game)
 
         raise Exception("Unknown procedure")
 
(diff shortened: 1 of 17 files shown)


danj3000:
I've added a number of changes that follow on from the reroll procedure approach. 

I've added it to the Block procedure as this is the hardest / most complex scenario to resolve - you have the choice of a number of different dice outcomes as well as the reroll / don't reroll options, plus the possibility of the opponent having the choice...

ReRoll is probably a misnomer at this point - It's a dice-result-chooser, with the added choices of re-roll.

I've added unit tests for the most common scenarios (not uphill), and have tested the uphill cases using the UI. Also have run a grodbot vs grodbot match through the UI, and it seems pretty happy - it even chose to reroll a block roll when it had 3 unfavourable dice on offer. I have added code to grodbot to never reroll a Loner failure.

remaining work would be to apply the reroll proc to the remaining procedures like catch, gfi etc, but these are mostly straightforward cases. This can be driven by writing unit tests to prove Loner applies in each case.

After that we can add Pro and Leader to the ReRoll proc so that they apply in all reroll situations with minimal effort.

Issue #87, opened 2019-12-11
Saving a game doesn't work


danj3000:
Error message is displayed.

danj3000:
The immediate problem with file references is easily fixed, but then the game will disqualify the team on reload.

Removing the clocks at save prevents this, but then how to re-establish the clocks on restart ??

njustesen:
Should probably be connected to a user management system. I.e. users has saved games.

njustesen:
Niels will do a fix without forward model first.

PR #93, opened 2019-12-30
Save game / resume fixes

danj3000:
Fixing issue #87 
part 1: A bug with absolute file paths.
part 2: pause clocks on save, then restart clocks on resume.

I have added a clock.resume to the check_clocks method. Not 100% sure if this will cause any side effects?


njustesen:
I wasn't able to load a saved game with this solution. I think a proper saveD games need to be connected to users, which we don't have atm. I will remove the "next release" tag from the issue as I don't think this is important right now.

njustesen:
I will merge these changes anyway since it is an improvement.

PR #94, opened 2020-01-12
Decay implementation plus tests

danj3000:
as title


njustesen:
This has been merged since I already had Decay implemented but without tests.

PR #106, opened 2020-04-18
Bot high kick fix

Christy42:


njustesen:
The fix only seemed to work for the scripted bot and not Grodbot. I changed the code a bit and it seems to work. Please review the changes in the HighKick function in Grodbot.

PR #113, opened 2020-07-15
Bribe and foul

mrbermell:
This pull requests adds automatic use of bribes if there are any. It's prepared for adding the action "USE_BRIBE" and "DONT_USE_BRIBE" but I didn't want to make such big steps yet. 

(This branch used to contain commits from my pull request for the pass skill fix, I reverted them to keep the pull requests to prevent overlap).


njustesen:
Thanks for this. I have implemented this feature using some of your code.

PR #114, opened 2020-08-11
Hypnotic gaze

mrbermell:
Implements skill hypnotic gaze in procedures.py and a test. 

Missing full web implementation still: 
* Web icon for hypnotized is currently same as really stupid. 
* Missing roll indication when the action is available (e.g. "3+" over the opponents square)
* would like to imitate leap's behavior in web view
* Player state might not properly be shown as "Hypnotized" when a hypnotized player is hovered.


mrbermell:
I noticed a strange behavior that a hypnotized player was not reset properly after a TD. Need to investigate some more..

njustesen:
> I noticed a strange behavior that a hypnotized player was not reset properly after a TD. Need to investigate some more..

PlayerState.reset() is called during the ClearBoard procedure. Here, it sets self.hypnotized = False, so it should work since that's what you are using.

A suggestion to improve the implementation is to make to new functions in Game. One that returns eligible hypno targets of a player and on the returns the roll target, it would reduce the followin code a lot - and you would also get tid of some code in the hypno procedure.

```
for opponent in self.game.get_adjacent_opponents(self.player, down=False):
    if opponent.has_tackle_zone(): 
                   hypno_positions.append(opponent.position)

                    modifiers = 1 - self.game.num_tackle_zones_in(self.player) 
                    target = Rules.agility_table[self.player.get_ag()]
                    agi_rolls.append([min(6, max(2, target - modifiers))])
```

mrbermell:
Regarding PlayerState.reset(), I found that quickly afterwards, but nevertheless I couldn't explain why on the web server that player seemed to remain hypnotized. I got a high kick too so I picked him for to step under the ball and he didn't even roll a catch dice. Was able to activate him afterwards though. I'll try to reproduce and see what's going on. 

Yes, your suggestion with functions in Game sounds very appropriate! 

Thanks for the feedback :)

mrbermell:
I've moved the logic to game.py. Can you have a look and say if it was according to your expectation?  Also have a look at the test at tell me if it's missing something or doing something stupid :) 

I think it's best if leave the web integration to you if that's ok. Still strange things going on with  showing the roll.. 

Later this week I'll add the vampire roaster and start bloodlust when this is finished.

njustesen:
Sounds good! I will take a look soon.

njustesen:
The changes look good. 
Can you please merge in the new changes in master? I made a quick attempt but wasn't sure if I overrode something you added.
I'll take care of the web app afterward.

mrbermell:
Yeah,I'll sort that out!

mrbermell:
Done. I had (accidentally?) merged the foul and bribe branch to this branch. A simple revert did the trick and the tables were straight forward to merge. Seems to pass all the tests in ffai/tests too.

mrbermell:
Branch has been merged.

PR #121, opened 2020-09-20
Blood Lust skill with tests

mrbermell:
At the time of writing this PR is for reviewing. It's not completely ready. Pending implementation of web and the opportunity to change block action into move action if BL roll fails.


njustesen:
I finished the implementation and it seems to work. 
It probably needs more testing and a test for the change from block to block or move.

mrbermell:
I'll play some vampire game and see how it works! And I'll propose a solution for the change move to block mechanic.

PR #123, opened 2020-10-03
Bugfix - apothecary condition and safe throw implementation 

mrbermell:
Safe throw was implemented according so that a fumbled outcome was treated as inaccurate. This is wrong according to BB2016. This PR solves that and also fixes an obvious error regarding apothecary in the casualty procedure. 

image

Sorry for the span with the previous PR that ended up being wrong.


njustesen:
Thanks! Great fix. 

I manually merged the changes in this commit https://github.com/njustesen/ffai/commit/9d3782b52a419a8eba9a6024de3a53537bf14d2f.

PR #127, opened 2020-10-15
Strip Ball and TIMMM-BER! 

mrbermell:
I took the conservative decision and named it "TIMBER" in the code.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -131,7 +131,7 @@ A race is supported if all positions can be used and all their starting skills a
 | Stab              | YES         | NO        |
 | Stakes            | YES         | NO        |
 | Stand Firm        | YES         | NO        |
-| Strip Ball        | NO          | NO        |
+| **Strip Ball**    | YES         | YES       |
 | **Strong Arm**    | YES         | YES       |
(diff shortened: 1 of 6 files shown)


njustesen:
I changed 
`min(2, len( [p for p in self.get_adjacent_teammates(player, down=False) if self.num_tackle_zones_in(p)==0 ] ) ) `
in `get_stand_up_modifier` to
`len([p for p in self.get_adjacent_teammates(player, down=False) if self.num_tackle_zones_in(p) == 0])`
because I couldn't see why. Tests still pass.

mrbermell:
Since you always fail on a natural one in practice you can only get support from two teammates bringing the roll to a 2+. That's why it was there. 

The procedure is currently not checking if the dice rolls is natural one (because I'm lazy). So right now it's a bug. 

Do you prefer to check for natural one in the procedure?

njustesen:
Dice rolls by default fail on lowest possible value and succeed on highest value, so even if three assisting players will make it a 1+ roll, a roll on 1 will fail. 

This is the same flow as other dice rolls, so I'll keep it like this.

PR #128, opened 2020-11-09
Performance increase in layers.py

mrbermell:
1) An abstract class "MemoizedFeatureLayer" to speed up generating certain layers. 
2) Some optimized layer generation methods
3) Small test for the profiler. 

20% speed increase in the profiler test on my laptop! 

possibly need to add cProfile as a package to the requirements now.

### DIFF
--- a/ffai/ai/layers.py
+++ b/ffai/ai/layers.py
@@ -17,14 +17,51 @@ class FeatureLayer:
     def name(self):
         raise NotImplementedError("Must be overridden by subclass")
 
+class MemoizedFeatureLayer(FeatureLayer):
+    def __init__(self): 
+        self.original_produce = self.produce 
+        self.produce = self.produce_memoized 
+        self.memory = {} 
+    
+    def produce(self, game):
+        raise NotImplementedError("Must be overridden by subclass")
+
+    def name(self):
+        raise NotImplementedError("Must be overridden by subclass")
+    
+    def calculate_key(self, game): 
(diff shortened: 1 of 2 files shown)


njustesen:
I like this idea. Also used it on the CrowdLayer.
I made the option to cache layers part of the FeatureLayer class thus removing the MemoizedFeatureLayer class. Haven't done the performance tests but it should have the exact same functions.

Ideas for future improvements:

1. Let FeatureLayers listen to events (outcomes) and possibly procedures added to the stack. Only re-produce layers if they appeared. This is possibly error-prone where some layers wouldn't be reproduced correctly. It would need tests for each layer. Not sure the effort would be worth it, but it would be fast.
2. Loop through all squares on the pitch ones, and call each feature layer per square.

mrbermell:
Great! 

(1) I have a hard time seeing which layers could benefit from this. Are there more than TargetPlayerLayer and BlockDiceLayer? 

(2) I like this idea but I'd like to shift it from squares to players. Not all but most of the Layers are centered around a player (used, down, stunned, MA, ST, etc..), so let's define way separate part that handles this. Then you only need to loop though all players on the pitch once. In practice the layer's produce function would a Player-object as argument instead of a Game-object. 

(3 new) I think it could also be worthwhile investigating what other things happening to the observations after they've been generated. There's some stacking and pytorch action going on in the a2c_example.py.

PR #131, opened 2021-02-27
Forward model

mrbermell:
Ready for a detailed review.


njustesen:
I had to change the flat monte carlo test to this to get it to run a game correctly (it's a new script now - not a test):

```
import ffai
from ffai.core import Action, Agent
import numpy as np
from copy import deepcopy
import random


class MCTSNode:
    def __init__(self, action):
        self.action = action
        self.evaluations = []

    def visits(self):
        return len(self.evaluations)

    def visit(self, score):
        self.evaluations.append(score)

    def score(self):
        return np.average(self.evaluations)


class FlatMCTSBot(ffai.Agent):

    def __init__(self, name, seed=None):
        super().__init__(name)
        self.my_team = None
        self.rnd = np.random.RandomState(seed)

    def new_game(self, game, team):
        self.my_team = team

    def act(self, game):

        print(f"Turn {self.my_team.state.turn}")

        nodes = []
        game_copy = deepcopy(game)
        game_copy.home_agent.human = True
        game_copy.away_agent.human = True

        root = game_copy.get_forward_model_current_step()

        for action_choice in game_copy.get_available_actions():
            if action_choice.action_type == ffai.ActionType.PLACE_PLAYER:
                continue
            for player in action_choice.players:
                nodes.append(MCTSNode(Action(action_choice.action_type, player=player)))
            for position in action_choice.positions:
                nodes.append(MCTSNode(Action(action_choice.action_type, position=position)))
            if len(action_choice.players) == len(action_choice.positions) == 0:
                nodes.append(MCTSNode(Action(action_choice.action_type)))

        best_node = None
        print(f"Evaluating {len(nodes)} nodes")
        for node in nodes:
            game_copy.step(node.action)
            score = self._evaluate(game)
            node.visit(score)
            # print(f"{node.action.action_type}: {node.score()}")
            if best_node is None or node.score() > best_node.score():
                best_node = node

            game_copy.revert_state(root)

        print(f"{best_node.action.action_type} selected.")

        return best_node.action

    def _evaluate(self, game):
        return random.random()

    def end_game(self, game):
        pass


# Register the bot to the framework
ffai.register_bot('mcts_bot', FlatMCTSBot)

# Load configurations, rules, arena and teams
config = ffai.load_config("bot-bowl-ii")
config.competition_mode = False
ruleset = ffai.load_rule_set(config.ruleset)
arena = ffai.load_arena(config.arena)
home = ffai.load_team_by_filename("human", ruleset)
away = ffai.load_team_by_filename("human", ruleset)
config.competition_mode = False
config.debug_mode = False
config.fast_mode = True

# Play 10 games
for i in range(10):
    bot_a = ffai.make_bot("mcts_bot")
    bot_b = ffai.make_bot("mcts_bot")

    game = ffai.Game(i, home, away, bot_a, bot_b, config, arena=arena, ruleset=ruleset)

    print("Starting game", (i + 1))
    game.enable_forward_model()
    game.init()
    print("Game is over")
```

**Two times though, I have seen this error:**
```
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/examples/monte_carlo_example.py", line 100, in <module>
    game.init()
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/ffai/core/game.py", line 100, in init
    self.step(start_action)
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/ffai/core/game.py", line 140, in step
    self.action = self._safe_act()
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/ffai/core/game.py", line 277, in _safe_act
    action = self.actor.act(self)
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/examples/monte_carlo_example.py", line 64, in act
    game_copy.revert_state(root)
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/ffai/core/game.py", line 74, in revert_state
    self.state_log.step_backward(to_step)
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/ffai/core/forward_model.py", line 67, in step_backward
    log_entry.undo()
  File "/Users/noju/git/tmp/ffai-fm-2/ffai/ffai/core/forward_model.py", line 113, in undo
    self.backward_func(self.owner, *self.backward_args)
KeyError: <Skill.DODGE: 9>
```

njustesen:
It would be cool if we could somehow hideaway this stuff but I don't really know how. 
It won't be clear to most people why you need to do it.
```
game_copy.home_agent.human = True
game_copy.away_agent.human = True
```

mrbermell:
I've tracked down the KeyError, something is wrong with the LoggedSet class. The undo() function is trying to remove Dodge from the used_skill property in player but fails because it's empty.  I'll add proper unit tests for LoggedSet and take it from there. 

Regarding the MCTS-test I added. I think it's strange to have such a specific test, I'll rewrite to make it more readable and more to the point. 

I'm currently tracking down a bug that happens when reverting the CoinTossProcedure and then comparing the states. The compare fails because it's checking for identity when it should compare attirbute equality. I think it's enough to re-write the compare functions but it's taking some time because of the very many different models and their nestled structure.

mrbermell:
> It would be cool if we could somehow hideaway this stuff but I don't really know how.
> It won't be clear to most people why you need to do it.
> 
> ```
> game_copy.home_agent.human = True
> game_copy.away_agent.human = True
> ```

Best way I can come up with is adding a function in Game that modifies the behavior to something between fast_mode=True and fast_mode=False.  It steps until it needs a action from any actor.  That function could either modify the agents as above or be some additional code in Game.step(). I don't like the first option because it'd be a very unexpected side effect of the function.

njustesen:
- Used skills are cleared every turn, maybe it's not tracking the clearing?
- The Flat MCTS script will eventually become an example script and part of a tutorial so it makes sense to have a much simpler test
- Something like fast_mode makes sense, thanks. I am starting to think that Game should never call agents. It should only perform one step and then wait. Something at a higher-level should control the agents. This is, however, not for now :)

mrbermell:
- I've just pushed an update that solves all the known bugs. However, all procedures are definently not tested since the tests plays only random games and revert them, when I have time I'll make more extensive tests 
- I agree with your point about the Game class. Also I think it has grown into a beast and needs splitting up. My spontaneous idea is to split it into three parts (but I've not given this enough thought): 
  - All getters that contain blood bowl rules e.g. get_pass_modifiers() 
  - All setters that manipulate the state e.g. put(), move(). And all getters that don't know any blood bowl rules e.g. get_player_at() 
  - The rest, which I think mainly is clocks and step() with its related subfunctions. 

You were gonna think more about appropriate names for my classes. Did you reach any conclusion? Let me know and I'll refactor!

njustesen:
There seems to be a problem with reverting unless I misunderstand how it's working:

Run this:
```python
def test_steps():
    game = get_game()
    game.home_agent = Agent("Human 1", human=True, agent_id=1)
    game.away_agent = Agent("Human 2", human=True, agent_id=2)
    game.init()
    game.enable_forward_model()
    game.config.fast_mode = True

    def avail_actions_str(game):
        return "-".join([action_choice.action_type.name for action_choice in game.state.available_actions])
    
    while not game.state.game_over:
        tmp_step = game.get_forward_model_current_step()
        tmp_available_actions_str = avail_actions_str(game)
        while True:
            action_choice = game.rnd.choice(game.state.available_actions)
            print(action_choice.action_type.name)
            if action_choice.action_type != ffai.ActionType.PLACE_PLAYER:
                break
        position = game.rnd.choice(action_choice.positions) if len(action_choice.positions) > 0 else None
        player = game.rnd.choice(action_choice.players) if len(action_choice.players) > 0 else None
        action = ffai.Action(action_choice.action_type, position=position, player=player)
        game.step(action)
        game.revert_state(tmp_step)
        assert tmp_available_actions_str == avail_actions_str(game)
        game.step(action)
```

mrbermell:
Thanks for finding this. I've pushed a commit with your test and an update that solves the bugs.

ffai/core/forward_model.py
```diff
@@ -0,0 +1,359 @@
...
+        new_value = new_type(value)
+        new_value.set_logger(logger)
+        return new_value
+    else:
+        raise AttributeError(f"Unable to add logging to {value}")
+
+
+class Immutable:
+    def __setattr__(self, key, value):
+        if hasattr(self, key):
```
mrbermell, line 356: Immutable's __setattr__ is responsible for a lot of computation time according to the cProfiler.

mrbermell (commented):
Performance issues with making Square immutable needs to be solved at some point

njustesen:
This makes the slow test much faster:
```python
def test_forward_model_revert_few_steps():
    game = get_game(fast_mode=True, human_agents=True)
    game.config.pathfinding_enabled = False

    def avail_actions_str(game):
        return "-".join([action_choice.action_type.name for action_choice in game.state.available_actions])

    tmp_step = game.get_forward_model_current_step()
    tmp_available_actions_str = avail_actions_str(game)

    while not game.state.game_over:
        t = time.time()
        game.step(get_random_action(game))
        game.revert_state(tmp_step)
        print("Reverted in: ", time.time() - t)
        assert tmp_available_actions_str == avail_actions_str(game)
        game.step(get_random_action(game))
        tmp_step = game.get_forward_model_current_step()
        tmp_available_actions_str = avail_actions_str(game)
```

njustesen:
The test fails when comparing the stack content:
```python
def test_forward_model_revert_few_steps():
    game = get_game(fast_mode=True, human_agents=True)
    game.config.pathfinding_enabled = False

    def avail_actions_str(game):
        return "-".join([action_choice.action_type.name for action_choice in game.state.available_actions])

    def stack_str(game):
        return "-".join([type(proc).__name__ for proc in game.state.stack.items])

    tmp_step = game.get_forward_model_current_step()
    tmp_available_actions_str = avail_actions_str(game)
    tmp_stack_str = stack_str(game)

    while not game.state.game_over:
        game.step(get_random_action(game))
        game.revert_state(tmp_step)
        assert tmp_available_actions_str == avail_actions_str(game)
        assert tmp_stack_str == stack_str(game)
        game.step(get_random_action(game))
        tmp_step = game.get_forward_model_current_step()
        tmp_available_actions_str = avail_actions_str(game)
```

PR #134, opened 2021-03-25
A2C example speed up and readability improvement

mrbermell:
* Moves the RL agents act() from the main process to the workers
* The workers finish a complete episode, then feedback the training data back to the main process. 
* Readability and modularity with new classes for different layers of abstraction. 

If the general idea of this code is approved then I'll add plots, better reports and a tutorial outline (and chase down some more bugs!).


njustesen:
I think it looks good so far.

From what I can read, async. workers push to a centralized memory buffer and the environment waits until it's full before returning training data back to the agent. If that's correct, it makes sense, and it's a small change to the A2C implementation. It's perhaps a misuse of the term environment because it's no longer taking a step. Maybe you can make the code simpler by not using the environment (including step()) terminology?

A small thing: In FFAI I've used the convention of calling functions `num_somethings()` to refer to the number of things. Please use that convention. 

Let's wait until you can confirm, perhaps through a tutorial (just add one), that it's faster than A2C - perhaps for the first three environments. The direction of this PR looks good overall.

mrbermell:
Your summary is correct (if I was picky I'd say the environment gives the training data to the optimizer and not the agent ☺️). Glad you like it! 

Thanks for the feedback about naming. I agree with all. I'll make a suggestion in the next commit. 

About performance, no use to have this unless it's actually better. Is it only total training time to beat random bot that counts? I can also think of:  
* Total number of time steps and/or optimizer steps to beat random bot (algorithmic performance)
* Average to generate X number of time steps of training data (i.e. without the optimizer step) 

Is it fair to use the default hyperparameters in A2C for all environments sizes? 

I'll make sure to enable pathfinding too!

njustesen:
>  Is it only total training time to beat random bot that counts?

You want to compare the algorithms on two things: 1) Wall-time clock, and 2) total environment steps, in which it can reach the same or better win rate. It's okay if it's better on (1) but not on (2), but ideally it's better on both. It's not okay (but interesting) if it's better on (2) but not (1).

> Is it fair to use the default hyperparameters in A2C for all environments sizes?

As long as you compare the two algorithms, it should be fine to keep them constant across the environments.

mrbermell:
Preliminary results are in for my algorithm (which I call A3c but it's completely as described in the paper). I'm currently patching the a2c example code to get compatible plots. 

a3c_3v3_preliminary

Figures on the first row have timesteps on their x-axis. Second row figures have wall clock time. 

image

Sample efficiency of pathfinding seems to be better with this algorithm. Hard to say anything about wall clock time.

njustesen:
This looks promising! Will you also run the original A2C baselines? I can also try to find my log files if you want to plot them.

mrbermell:
Right now it's fair to say that a3c in it's current configuration doesn't hold up to the a2c performance. I'll tweak it according to disucssions we've had on Discord. More results soon!

mrbermell:
I'm closing this PR because lack of results. There could still be interesting to go in this direction but I will not focus on it right now.

PR #138, opened 2021-05-07
removes execute flag from all files

derNarr:
all files had the execute bit set, which seems unnecessary and is (at least for me) annoying. Therefore I removed it from all files with ``chmod -R -x+X *`` .


njustesen:
Thanks. This seems reasonable but some files, especially those inside examples/, are meant to be executed and I guess some people prefer:
`./myscript.py` instead of `python myscript.py`, which is not possible with these changes.

If you want to go through the files and only remove the execute flag on those without a main, I can accept that. 

If it's best practice to remove the flag on all files, please provide a link so I can read up on it.

derNarr:
In the examples on the website it is suggested to build a conda environment. Therefore the environment should be activated before execution anyway. I agree that running ``./myscript.py`` does not work in a Terminal with a correctly activated environment. On the other hand, you are not prompted to execute the script, when double clicking, which is (nearly) guarantied to fail, as this will usually be executed in the systems python environment.

All examples on the website use the ``python myscript.py`` way of running scripts anyway. Personally, I reserve the execution bit for things I run regularly as a program.

I could not find any good link on the topic yet, but I have never encountered python files (or all files) to be executable, when cloning a python project, except for those in a ``bin/`` folder.

Looking into ``numpy`` the ``setup.py`` and the ``runtest.py`` are executable. Additionally: ``doc/postprocess.py`` and some of the files (but not all) in the ``tools/`` folder.

Looking into ``pytorch``: ``setup.py`` is not exectable, but most shell scripts are and some python files are, but most are not. Superficially, I could not spot a pattern.

If you feel like I should make all files with a ``if __name__ == "__main__":`` executable, let me know, I still feel it would be less confusing and more explicit to remove all execution bits as it forces the user to run ``python myscript.py`` in a Terminal, which makes it easier to debug and to check if the proper environment is activated.

njustesen:
Closing this. I am willing to open this again if more people see this as a problem.

PR #141, opened 2021-05-17
Bugfix for observations out of range, and test to verify

mrbermell:


njustesen:
Great. Looks good.

I believe accumulated_prob_2d_roll can be optimized by making the array static and as a Numpy array. 

We should run A2C on the 1v1 variant to verify that we didn't break anything. Will you do that or should I?

ffai/ai/layers.py
```diff
@@ -313,9 +320,13 @@ def produce(self, game):
                         out[action_choice.positions[i].y][action_choice.positions[i].x] = action_choice.paths[i].prob
                     elif i < len(action_choice.rolls):
                         # Convert to chance of succeeding
-                        chance = 1.0
-                        for roll in action_choice.rolls[i]:
-                            chance = chance * ((1+(6-roll)) / 6)
+                        if action_choice.action_type == ActionType.FOUL:
+                            # Use different probability calculation for 2D6 rolls
+                            chance = self.accumulated_prob_2d_roll(action_choice.rolls[0][0])
```
mrbermell, line 325: First zero in the index should be "i", I think.

ffai/ai/layers.py
```diff
@@ -592,8 +581,27 @@ def produce(self, game):
...
+class GFIsLeftLayer(FeatureLayer):
+
+    def produce(self, game):
+        out = np.zeros((game.arena.height, game.arena.width))
+        active_team = game.state.available_actions[0].team if len(game.state.available_actions) > 0 else None
+        if active_team is None:
+            return out
+        for player in active_team.players + game.get_opp_team(active_team).players:
+            if player.position is not None:
+                has_sprint = player.has_skill(Skill.SPRINT)
```
mrbermell, line 600: Rewrite as: num_max_gfi = 2 if... else 3. To make be able to write next line without if statements.

ffai/ai/layers.py
njustesen: Yes. chance = chance * ((1+6-roll) / 6) looks a bit cleaner.

mrbermell:
I'll make the array static, fix two other small things and then test the a2c example!

mrbermell:
I've updated according to agreement. Picture below from A2C for 1v1 with standard parameters. Seems to work fine. 

image

mrbermell:
I'm closing and re-creating this PR. I accidentally merge another branch into it and I can't get the PR to revert the commits. Sorry for the hazzle..

PR #142, opened 2021-05-26
Game class now exposes max number of rounds in json as game.rounds an…

derNarr:
…d this is used in js controller to show the maximal number of rounds in the web interface

This PR exposes the number of rounds per half in the json and uses this information to display it correctly in the web interface.

Git was complaining about not having a newline at the end of ``static/js/controllers.js`` so I added one, but I am not sure, if this is good or not.


njustesen:
This looks good. I'll pull it down and run a test but it should be ready to be accepted.

PR #143, opened 2021-05-26
fix formation setup code, if line of scrimmage is not the first line

derNarr:
when the line of scrimmage was not on the first line of the field, the code parsing the formation files showed strange behaviours. This change fixes these, by looping over all positions and checking, if a position is on the line of scrimmage and not assuming that the LoS is the first line. The old code assumed that the LoS is the first line.

### DIFF
--- a/ffai/core/model.py
+++ b/ffai/core/model.py
@@ -1227,25 +1227,26 @@ class Formation(Immutable):
 
         positions_used = []
 
-        # setup on scrimmage
+        # first setup on scrimmage
         for t in ['S', 's', 'p', 'b', 'c', 'm', 'a', 'v', 'd', '0', 'x']:
             for y in range(len(self.formation)):
-                if len(players) == 0:
-                    return actions
-                x = len(self.formation[0])-1
-                tp = self.formation[y][x]
-                if tp == '-' or tp != t:
-                    continue
-                yy = y + 1
-                xx = x + 1 if not home else game.arena.width - x - 2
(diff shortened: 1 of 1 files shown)


njustesen:
Please add a test for this. You could also consider making one big BB Sevens PR.

derNarr:
I will add a test into `tests/kickoff/test_setup.py` .

I will create a pull request for threes and sevens, when I feel they are only some well defined tasks left.

If you don't mind, I would continue to cherry pick the general improvements of the code base and create separate pull requests on the way. This makes the final pull request of sevens and / or threes less painful.

The main showstopper for merging threes and sevens at the moment is, that I don't know yet how to switch between the different game modes. At the moment I change the `server.py` and the `web.json` file to switch between game modes in the front end.

If you would prefer one big pull request in the end, I can do that as well. You are the maintainer and author anyways :-)

njustesen:
Small PRs are fine! It would be cool to switch between game modes through the UI.

Any progress on this?

derNarr:
I have not worked on it for a while. The sevens pitch works in principle but need some polishing. I have no real clue on how to change the UI substantially, like adding a new drop down menu with the game mode and then acknowledge that in the python code. I do feel comfortable with the python code base though.

There are more general code changes in my sevens branch, which I might bring into small PRs, if these are merged. :-)

njustesen:
Cool. I know my way around the UI so I can easily do that.

PR #151, opened 2021-06-10
Fix frenzy stab

derNarr:
This is a first attempt to allow for a stab after a block on a frenzy player.

Tests go through smoothly now, but I have to do some manual testing with this as well.

Code for Blitz is still missing.

(This PR also contains the fix for the throw teammate as I branched from the wrong starting branch -- I might fix this, if the throw teammate PR is not merged soon.)

### DIFF
--- a/ffai/core/game.py
+++ b/ffai/core/game.py
@@ -2397,7 +2397,7 @@ class Game:
                                         positions=positions, rolls=d6_rolls))
         return actions
 
-    def get_block_actions(self, player, blitz=False):
+    def get_block_actions(self, player, blitz=False, defender=None):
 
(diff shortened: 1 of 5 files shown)


derNarr:
This one is not correct yet. It allows to stab anyone after the first Frenzy DO NOT MERGE :-)

njustesen:
What's the status on this one? Seems like good work but not quite there?

derNarr:
I did not cracked it, but now after some time is gone by, I might give it a fresh look. I let you know, if I make any progress. As this was a TODO in the code, I thought I share my still unsuccessful steps in this direction.

PR #152, opened 2021-06-24
Forward model bugfix - game.actor not properly reset

mrbermell:
game.actor should now be properly reset after calling game.revert()


njustesen:
Can you try to pull from master and commit again? It should now run the tests on pull requests as well.

PR #153, opened 2021-07-30
[Proposal] Enable deterministic tree search 

mrbermell:
Adding code to show example based on discussion on discord. Only for discussion. Not for merge!

### DIFF
--- a/ffai/core/model.py
+++ b/ffai/core/model.py
@@ -564,7 +564,7 @@ class DiceRoll(Reversible):
     def __init__(self, dice, modifiers=0, target=None, d68=False, roll_type=RollType.AGILITY_ROLL, target_higher=True, target_lower=False, highest_succeed=True, lowest_fail=True):
         super().__init__()
         self.dice = dice
-        self.sum = 0
+        self._sum = None 
         self.d68 = d68
         self.target = target
         self.modifiers = modifiers
@@ -573,14 +573,7 @@ class DiceRoll(Reversible):
         self.target_lower = target_lower
         self.highest_succeed = highest_succeed
         self.lowest_fail = lowest_fail
-        # Roll dice
-        for d in self.dice:
-            if not isinstance(d, BBDie):
-                if d68 and isinstance(d, D6):
-                    self.sum += d.get_value() * 10
-                else:
-                    self.sum += d.get_value()
-
+        
     def to_json(self):
         dice = []
         for die in self.dice:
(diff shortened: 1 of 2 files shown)


njustesen:
Just a quick comment before diving into the code. Can't you use the dice rolls information in the available actions for this? And then use D6.fix().

mrbermell:
I did consider this. And of course it works for movement actions and ball handling. But not for some special cases e.g. ball bounce and throw in, armor/injury/Cas rolls, kickoff event. 

One could of course hard code the important special cases to start with. Maybe that saves us a lot of work and there is more high yield features to work with. What do you think?

njustesen:
I like the idea but I think procedures should be kept as simple as possible. 

Let's maybe start with the interface you want, then it's easier to discuss implementations. Could you write the function signature you want? I imagine something like this:


```python
def expand(game, action):
     """Returns a list of tuples containing (trajectory, probability) for each possible outcome"""
```

mrbermell:
I didn't think about the higher level abstractions yet when I posed the question but it's a good idea to start from the top. The forward model is used but no need for a complete Trajectory, simply a list of ffai.core.forward_model.Step that can be appended to Trajectory object of the Game object. Here's how I imagine it's used. 

```python
from ffai.core.forward_model import Step

Steps = list[Step]  # type alias

def expand(game, action) -> list[Tuple[Steps, float]] 
     """Returns a list of tuples containing (Steps, probability) for each possible outcome"""
    pass 

def calculate_heuristic(game) -> float: 
    """Returns heuristic for home team based on state of game  """
    pass 

def calculate_average heuristic(game, action) -> float: 
    """Returns heuristic for home team based on all possible outcomes from action. Leaves game in original state.  """
    heuristic = 0     
    root = game.get_step() 
    for steps, prob in expand(game, action): 
        game.trajectory.action_log.append(steps)  
        game.trajectory.step_forward(len(game.trajectory.action_log))  # Trajectory.step_forward is currently not implemented. 
        heuristic += prob * calculate_heuristic(game) 
        game.revert(root) 
    return heuristic 

```

njustesen:
Looks good. I usually use raise NotImplementedException instead of pass if it's meant to be overridden.

What's the status on this one?

mrbermell:
I haven't worked further with this. I got stuck trying to re-write the pathfinding with cython/C++ 🤓 as soon as I'm done I will look further into this!

PR #156, opened 2021-09-03
Pathfinding does not stop pre-maturely if no moves are left

### LINKED ISSUE #155: Player turn ends automatically during handoffs and fouls when no more moves are left
This happens only when pathfinding is enabled.

njustesen:
Fixes #155


mrbermell (approved):
Looks good :)

PR #160, opened 2021-09-12
2020 ruleset delta

mrbermell:
Finalized the list of changes to the best of my knowledge.


njustesen (approved):
Looks great! My only comment is avoid having spaces in filenames :)

We could also think about using the checkbox syntax that GitHub supports to track what's implemented. It's [ ] and [X] and allows you to click to change.

mrbermell:
Great :) I will change the filename to something more appropriate and add the checkboxes. 

What do you think about supporting two different rulesets? Is it reasonable to try? I fear procedure.py and game.py will get bloated down if-statements checking the rules.

njustesen:
> I fear procedure.py and game.py will get bloated down if-statements checking the rules.

If to many if-statements are added to make a procedure support both rulesets it might be easier to split it into two procedures.

What would be a good feature to start off with that would challenge our architecture?

mrbermell:
> What would be a good feature to start off with that would challenge our architecture? Maybe passing?

Yes, passing and/or the new way AG rolls work (having e.g. 3+ in the roaster).  If we try to implement everything that concerns throwing/catching the ball and some of the relevant skills, then we have a good challenge for the architecture!

njustesen:
Good job! We should maybe merge this with docs/features.md at some point.

PR #163, opened 2021-09-24
Pathfinding fix to limit only one foul and handoff per path

njustesen:
The pathfinder checked self.can_foul and self.can_handoff to see whether it could do a foul or handoff. This obviously needs to be tied the the state of the node so it's only allowed once. That's implemented in this PR.

Without this check, the pathfinder can actually end in a never-ending-loop.

### DIFF
--- a/ffai/ai/pathfinding.py
+++ b/ffai/ai/pathfinding.py
@@ -48,6 +48,9 @@ class Node:
     SURE_FEET = 2
     SURE_HANDS = 3
 
+    def __lt__(self, other):
+        return self.euclidean_distance < other.euclidean_distance
+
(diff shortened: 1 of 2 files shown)


mrbermell:
I'll review this weekend. Is the new foul test the one that will go one forever without this fix?

njustesen:
> Is the new foul test the one that will go one forever without this fix?

Yes. It was so hard to reproduce!

mrbermell (approved):
Looks good :) 
I suggest adding the same test for the handoff

njustesen:
Same test added for handoff. It's passing both with and without the fix - not sure why but that's my least concern :)

Issue #148, opened 2021-06-10
tests on gym hang occasionally


derNarr:
The tests in ``test_gym.py`` hang from time to time. This seems to be a result of the env setup code that sometimes leads to an empty list of available actions like::

```
ENV:  [<ActionType.START_GAME: 1>]
GAME:  START_GAME
ENV:  [<ActionType.SETUP_FORMATION_SPREAD: 54>, <ActionType.SETUP_FORMATION_ZONE: 53>]
GAME:  PLACE_PLAYER, END_SETUP, SETUP_FORMATION_SPREAD, SETUP_FORMATION_ZONE
ENV:  [<ActionType.PLACE_BALL: 9>]
GAME:  PLACE_BALL
ENV:  []
GAME:  PLACE_PLAYER, END_SETUP
```

In the example above after the ball was placed there is the option to place another player or end the setup. But as the environment code only allows to use `` SETUP_FORMATION_*`` the list of available actions is empty. (I changed the code to give only the intersection between available actions and actions allowed by the environment, the code in the main branch always gives the actions that are allowed by the environment, but these are illegal actions in this case.)

derNarr:
Ok, it seems to be related to the kickoff events. E. g. for ``KICKOFF_PERFECT_DEFENSE`` you are still in setup, but don't get the formations as available actions.

derNarr:
An easy fix would be to only expose the ``END_SETUP`` event when some kickoff event is rolled that allows for a change of the setup.

```
                formation_actions = [action.action_type for action in self.game.state.available_actions if action.action_type in self.offensive_formation_action_types]
                if formation_actions:
                    return formation_actions
                else:
                    return [ActionType.END_SETUP,]
```

njustesen:
I can't get the tests to fail even if I hard code the kickoff event to always be perfect defense. Can you write what you do as a test that always fails?

derNarr:
I changed the ``_one_step`` code to raise an Error for an action that is not available at the moment like:

```
    def _one_step(self, action):
        """
        Executes one step in the game if it is allowed.
        :param action: Action from agent. Can be None if no action is required.
        :return: True if game requires action or game is over, False if not
        """

        # Get proc
        proc = self.state.stack.peek()

        # If no action and action is required
        if action is None and len(self.state.available_actions) > 0:
            raise GameActionError(f"None action is not allowed when actions are available ({', '.join(aa.action_type.name for aa in self.state.available_actions)})")
            if self.config.debug_mode:
[...]

PR #166, opened 2021-10-01
Fix bug with perfect defense in gym (issue #148)

mrbermell:


njustesen (approved):
This is good. 

Some thought for the future: 
It would be nice to have the formations part of the configuration and not hardcoded into the engine.

We could have a ActionType.FORMATION and a way to point to the index in the list of formations added to the game. 

Alternatively, formations could be completely removed from the core and have it as a utility inside ffai.ai that FFAIEnv and bots can use.

PR #167, opened 2021-10-01
Exception on illegal action

mrbermell:
This PR replaces code in Game.py like: 
```python
if self.config.debug.mode: 
    print(f" <here's the error> ") 
```

with: 
```python
raise GameActionIllegal(f" <here's the error> ") 
```
This makes the code less forgiving, and bugs are easier to spot. The PR also solves some of the bugs that surfaced :smiley:


mrbermell:
I think it it should still raise the exception but the competition module should handle said exception and request a new action. I can write a test or two to check that it works with the competition. 

Agreed. I'll refactor the name!

njustesen:
We should write some tests for the competition actually.
Notice that the Competition class does not control the flow of the game so it can probably not be in charge of requesting the agent again.

mrbermell:
> Notice that the Competition class does not control the flow of the game so it can probably not be in charge of requesting the agent again. 

That makes it harder. Do you think it's reasonable to catch the exception in Game.step(). Though it's not code Uncle Bob would be proud of.
``` python 
# Perform game step
if self.dont_crash:  # or similar check of a config flag 
    try:
        done = self._one_step(self.action) 
    except InvalidActionError: 
        pass 
else:      
    done = self._one_step(self.action)
```

mrbermell:
Also, the CONTINUE action that is called by in the web UI at regular intervals which raises the exception every time `len(available_actions) > 0` which is quite often. I suggest simply remove that raised exception because it serves no purpose.

mrbermell:
> I agree, but when bots are submitted to bot bowl that behave in new and unexpected ways it will challenge the framework a lot and new errors might appear. We could go with your solution of just throwing the exception but that means during bot bowl that if the framework throws an exception, we have to fix the error and re-run the matchup - which is a big commitment. Calling the match a tie is the easy way out but obviously doesn't look good when reporting the results. Not sure what's best here. Make it configurable? :D

I don't think we need to be too worried about a bot breaking the framework since the interface to the bot is limited to Action-objects. I am more worried that a general exception clause would catch the exceptions should break the environment. But having a small error completely destroy the record of all matches that was success also sounds... well, not perfect. Especially if there are many tree searching bots that uses the maximum available time! Maybe a solution is to delegate each separate game or matchup to a pool of workers.  Or I just add the thing and we see how it goes. I don't know..

mrbermell:
> as unresponsive as they like without causing too much trouble. 

😆

Let's look into automating more parts of the comp during the spring!

PR #168, opened 2021-10-04
Documentation update and new structure for issues

mrbermell:
This PR contains some ideas for improving the documentation. With it I suggest we also start using labels for issues and connect them to milestones. We should aim to always have a few issues with label "good first issue" to encourage newcomers to get involved. We should also collect issues with label "research topic" to spark the interest of researchers and blood bowl playing students looking for thesis topics (though issues might not be the best medium for this). [Here's an example of how it could look](https://github.com/mrbermell/ffai/issues).


njustesen:
Looks great!

 My only comment is that it should be more clear that bot development and framework development isn't the same. New bot devs should be guided to the tutorials and framework devs to our dev docs. This should make it look easier to get started with bot development.

mrbermell:
Good point. I will make sure it is!

PR #169, opened 2021-10-05
Compiled pathfinding

mrbermell:
"[Cython](https://github.com/cython/cython) is a language that makes writing C extensions for Python as easy as Python itself."

I've translated the pathfinding algorithm to the cython syntax and configured a compilation. During the compilation, cython translates the code to pure C++ which is then compiled with a compiler you have available on your computer. I choose to focus on pathfinding because the profiler have identified this is biggest bottleneck. In a simple benchmark test I get 1/8 of the execution time compared to the regular python implementation. 

The compilation increases the installation threshold because it requires a C++ compiler. My intention with this PR is to only compile if the prerequisites are there, otherwise use the python implementation of the pathfinder. 

Another reason to keep the python implementation is that developing and debugging in python is much much easier compared to Cython. So algorithmic development should first be done in python and later adapted into Cython where large test suites can ensure that the two implementation behave identically. 

I now invite anyone and everyone to test this. With and without a C++ compiler available, because FFAI should work in both cases. 

To compile simply run this command in the root of the repository. 
```
python setup.py build 
```

Best of luck and let me know if you get stuck on anything. 

### Todo 
"setup.py build" works and all tests pass with... 
- [x] Cython and C/C++ compiler installed
- [x] Cython installed but no C/C++ compiler available
- [x] Without Cython and C/C++ compiler (not applicable, Cython can safely be added as a dependency for all users) 
- [x] Github checks
- [x] Refactor filenames and move to core/
- [x] Benchmark


njustesen:
I was able to install it on mac and if I am not fooling myself it feels faster to compute the paths for gutter runners.

We should strive to automate all of this as much as possible including copying the compiled .so. This feature is not only useful for RL but for anything since pathfinding is a core feature in the framework now.

njustesen:
I like that idea.

We could even do the import in `ffai/ai/pathfinding/__init__.py` to keep the same interface as now:

```python
try: 
    import foo, bar from compiled_pathfinding
except ImportError: 
    import foo, bar from python_pathfinding
```

mrbermell:
Merge done! :)

mrbermell:
Thanks for the feedback. I've made the updates you requested. I wasn't completely sure about point (2) though, let me know what you think!

njustesen:
I think this is ready to be merged!

PR #174, opened 2021-10-08
Correct action types on pathfinder-assisted actions

njustesen:
Seems to work. 

TODO:
- [X] Change to correct action types in pathfinder-assisted actions
- [ ] Restructure procedure code to be more readable
- [ ] Write some tests for this 
- [ ] Fix UI (missing cursor and path-steps on hover)

### DIFF
--- a/ffai/core/procedure.py
+++ b/ffai/core/procedure.py
@@ -2594,6 +2594,7 @@ class MoveAction(Procedure):
         self.player_action_type = player_action_type
         self.paths = {}
         self.steps = None
+        self.orig_action_type = None
         self.can_undo = self.game.get_team_agent(player.team).human
 
     def start(self):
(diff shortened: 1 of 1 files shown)


njustesen:
While fixing up the UI I accidentally made it show the block dice tip on all available block actions and not only the one it hovers over. I kinda like it. What do you think?

Screenshot 2021-10-10 at 23 08 58

mrbermell:
I think this is ready to merge. 

I found a few quirks, but not important enough to stop us from merging this PR in my opinion. I'll create a separate issue with them, here they are if you disagree. 
 - When activating Stab in the Web UI during a Block or Blitz action. Number of block dice should not be visible
 - When activating Stab in the Web UI during Blitz action. Show the relevant stab roll for opponents. This works when blocking but not when blitzing. 
 - Stab outcome reports off wrong roll target, one less than it should be. 
 - Proc bot and Scripted bot does not handle Shadowing procedure which freezes the game. (I admit not Stab-related).

PR #189, opened 2021-11-19
Improved installation docs regarding c++ compiler

### LINKED ISSUE #188: Improve documentation on how to build with cython when there's no c++ compiler
It's currently not straightforward to figure out how to fix:
```
You've built botbowl without cython compilation. Check docs/installation.md for details.
```
after running ```pip install .``` without a c++ compiler.

The docs/installation.md should provide some details instead of a link to the cython documentation.

njustesen:
Fixes #188 

Added some details and removed parts that no longer made sense.

E.g. it didn't say whether botbowl was installed with or without cython and the import check didn't work.


mrbermell:
Yes, this is necessary! Thanks for the help with cython docs!

Here are some of my findings/thoughts: 
Cython itself does not call the compiler, it just translates the python/cython code to c/c++ (depending on your choice). 
The compilation from c code to binary is done by setuptools.setup when provided an Extension object in the ext_modules argument: 
`setup(..., ext_modules=cythonize(...))` 

And how setuptools detects and identifies compilers is black magic to me, especially on windows.

mrbermell:
I got the compilation working in a Windows 10 virtual machine. Just installing [Visual Studio C++ Community](https://visualstudio.microsoft.com/vs/features/cplusplus/) was enough. Now I'm trying to to get the compiled executable copied into the correct directory. 

```python 
# from setup.py 

if compile_available:
    # Grab all compiled modules and copy into source folders
    compiled_file_type = ".pyd" if platform.system() == "Windows" else ".so"
    copied_files = 0
    for root, dirs, files in os.walk('./build/'):
        for file in files:
            if file.endswith(compiled_file_type):
                from_file = f"{root}/{file}"
                to_file = "./botbowl/" + root.split('/botbowl/')[1] + "/" + str(file)
                print(f"copying '{from_file}' -> '{to_file}'")
                shutil.copyfile(from_file, to_file)
                copied_files += 1
```

mrbermell:
I got the compilation working on Windows out of the box. I've made some updates in the documentation as well. I think it's ready to be merged. Please have a look and let me know what you think!

mrbermell:
Eventually I think we want to add: 
1. Test for windows integration in the Github Action workflow. 
2. Test the framework completely with and completely without compilation. It's possible to catch any custom arguments to the pip installer and can be accessed in setup.py. But it's somewhat tricky. 

Probably not very important right now but if more parts of the framework is compiled it becomes increasingly important.

mrbermell:
We should merge this :)

Issue #183, opened 2021-10-25
Add type annotations for functions (PEP 484)


mrbermell:
[PEP 484](https://www.python.org/dev/peps/pep-0484/)  specifies a syntax for providing type hints for function arguments and return types: 

``` python
def greeting(name: str) -> str:
    return 'Hello ' + name
```
This makes the code more readable and gives more information to linters which can point out inconsistencies. 

It would be great to have for all python files in the framework, especially the ones used for bot development. This is a great way to get started working with and understanding the framework. Any help is greatly appreciated! 

TODO: 
 - [x] core/game.py 
 - [x] core/model.py 
 - [ ] core/procedure.py 
 - [x] core/forward_model.py 
 - [ ] core/load.py 
 - [ ] core/table.py 
 - [ ] core/util.py 
 - [ ] ai/env.py 
 - [ ] ai/pathfinding.py 
 - [ ] ai/registry.py

mrbermell:
@phasenoisepon, great that you've taken on this! Do you need any help or do you have any questions?  =)

phasenoisepon:
Haven't had a lot of time to work on this, feel free to assign it to someone else.

mrbermell:
Ok, no worries!

PR #190, opened 2021-11-24
Annotations

mrbermell:
Starts work with #183 . 

solves a small bug with how python pathfinding interacts with the forward model. 

Adds __repr__() for some of the models.

### DIFF
--- a/botbowl/core/game.py
+++ b/botbowl/core/game.py
@@ -32,16 +32,17 @@ class Game:
     end_time: Optional[float]
     last_request_time: Optional[float]
     last_action_time: Optional[float]
-    forced_action: Any #??
     action: Optional[Action]
     trajectory: Trajectory
     square_shortcut: List[List[Square]]
 
-    def __init__(self, game_id, home_team: Team, away_team: Team,
-                 home_agent: Agent, away_agent: Agent,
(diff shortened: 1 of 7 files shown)


botbowl/core/game.py
```diff
@@ -5,17 +5,43 @@
...
+    config: Configuration
+    ruleset: RuleSet
+    state: GameState
+    rnd: np.random.RandomState
+    ff_map: Any #??
+    start_time: float
+    end_time: float
+    last_request_time: float
+    last_action_time: float
+    forced_action: Any #??
```
njustesen, line 35: There are no longer any usages of this.. we can probably remove it.

botbowl/core/game.py
```diff
@@ -5,17 +5,43 @@
...
+    action: Action
+    trajectory: Trajectory
+    square_shortcut: List[List['Square']]
+
+    def __init__(self, game_id, home_team: Team, away_team: Team,
+                 home_agent: Agent, away_agent: Agent,
+                 config: Optional[Configuration] = None,
+                 arena: Optional[TwoPlayerArena] = None, ruleset: Optional[RuleSet] = None,
```
njustesen, line 43: Add line break.

njustesen (approved):
Looks great! 

Only a few minor comments. Feels free to merge, 

Btw. have you checked if PyCharm gives any warnings now that the type annotations are added?
Code -> Inspect code

Issue #185, opened 2021-11-01
Out of the box RL friendly gym env


mrbermell:
There should be a gym environment that is intended for use with out of the box RL code from github and scientific papers. The current gym environment FFAI has requires a lot of adaptation (see a2c_example.py). 

By making an ffai gym env that is more integration friendly it becomes easier to benchmark state of the art algorithms for blood bowl and makes it easier to get started with the framework. 

Exactly how this could be done can be discussed, here are the two main issues: 

-  **Only spatial observations**. Move non-spatial observations into layers. This will create a huge spatial observation matrix, so maybe we need to question many of the none spatial observation. 
- **Actions as int**. This is the standard way of handling discrete action spaces and is already implemented in a2c_example.py 

The action mask needs to be handled somehow too, I think it's reasonable to have a separate getter-function for it instead of clogging down what is returned by gym.step() but I'm not sure...

njustesen:
I think you'll lose a lot by removing the non-spatial data stream. We are currently doing what DeepMind does in StarCraft and I believe OpenAI does for Dota, having a non-spatial and spatial input stream. The problem is that RL code from github and scientific papers are focused on small and simple environments so they cannot handle dual input streams - but they should!

mrbermell:
Yes, I know and I agree, they should! I'm not suggesting we scrap the idea of a non spatial input, just extend the framework. I think it's also worthwhile to have the ability to test that hypothesis: that non spatial input is needed. 

And to clarify, I'm not saying remove non spatial observations, I'm say add that information to spatial layers instead so the observation becomes a single tensor. Though it's an inefficient representation (to say the least!), it saves a lot of coding time. I don't want to know how many hours I poured into getting SEED RL work with multiple observation tensors...

njustesen:
I would suggest trying this out in a quick hacky way and if the results are good, then we can consider to building it into the gym env.

Issue #178, opened 2021-10-10
More formations


mrbermell:
Right now, ffai has four formations and it's complicated to add more. There should be more formations available and simple to add for more. 

Njustesen has suggested a new action called FORMATION, which has an additional attribute with a formation object. 

I think it's still reasonable to keep the default formations as separate actions but it should be more and some refactoring is probably due.

njustesen:
The aim is to move the logic out of the game class. 

Low priority.

njustesen:
Moving this out of the Bot Bowl IV Announcement milestone.

PR #197, opened 2021-12-29
New gym env

mrbermell:
This PR takes a stab at #185 and #178 by merging a2c example's functions into BotBowlEnv and managing formations in the gym Env. 

For now I've only added the new environment next to the old. 

Summary of changes and my thoughts: 

1. env.step() takes an int as argument
2. env.step() returns a tuple with  (spatial_obs, reward, done, info) where info is a dict that contains non_spatial_obs and action_mask. 
3. environment plays on both sides (home and away) as a opposed to the old BotBowlEnv that by default played against a RandomBot. Environment flips spatial observation and action when away team is playing. 
4. reward given once at end of game to home or away, whoever made the last action, -1 for a loss, 0 for a draw, 1 for a win. 
5. New formations can easily be added with one extra line of code specifying the path to the formation file. 

TODO: 
- [x] wrapper for shaped reward 
- [x] wrapper for scripted actions 
- [x] custom formations 
- [x] Update a2c tutorial
- [x] Re-run a2c_example.py and confirm results 
- [x] Fix renderer


njustesen:
Overall, some very nice steps towards a better gym environment. I think you should add to the list, that we want the tutorials to be updated too. 

Some comments:

> env.step() takes an int as argument

This makes sense but how does one figure out what the int means?

> env.step() returns a tuple with (spatial_obs, reward, done, info) where info is a dict that contains non_spatial_obs and action_mask.

Great!

> environment plays on both sides (home and away) as a opposed to the old BotBowlEnv that by default played against a RandomBot. 

So self-play is incorporated into the gym environment by default? I don't think we should do that because it makes the environment more complex and harder to compare results. There could be a switch (boolean in the constructor) that allows you to easily switch self-play on.

> Environment flips spatial observation and action when away team is playing.

I like the feature but I guess this is only needed for self-play so the user doesn't have to do the flipping?

> reward given once at end of game to home or away, whoever made the last action, -1 for a loss, 0 for a draw, 1 for a win.

Also only needed for self-play, I suppose. I think we should keep it simple.

> New formations can easily be added with one extra line of code specifying the path to the formation file.

That's cool, but as mentioned above, it would be nice if one could add the formation without altering the code. If EnvConfig had a constructor it would solve this.

mrbermell:
Thanks for the feedback 👍  Here are my responses: 

> > env.step() takes an int as argument
>
> This makes sense but how does one figure out what the int means?

I'm not sure I understand the question. The point is that the neural network learn how actions affects the environment? If you want to find out what the int means you can run `env._compute_action(action_idx)` which returns the corresponding Action object. 

> > environment plays on both sides (home and away) as a opposed to the old BotBowlEnv that by default played against a RandomBot.
> 
> So self-play is incorporated into the gym environment by default? I don't think we should do that because it makes the environment more complex and harder to compare results. There could be a switch (boolean in the constructor) that allows you to easily switch self-play on.

Make sense, I'll make it a game vs RandomBot as default and make a self-play switch in the constructor. 

> > Environment flips spatial observation and action when away team is playing.
> 
> I like the feature but I guess this is only needed for self-play so the user doesn't have to do the flipping?

Yes, only for self play. So it comes with the self play switch. 

> > reward given once at end of game to home or away, whoever made the last action, -1 for a loss, 0 for a draw, 1 for a win.
> 
> Also only needed for self-play, I suppose. I think we should keep it simple.

Come to think of it, the reward wrapper should handle all rewards so I'll move it there. 

> > New formations can easily be added with one extra line of code specifying the path to the formation file.
> 
> That's cool, but as mentioned above, it would be nice if one could add the formation without altering the code. If EnvConfig had a constructor it would solve this.

Ok, I'll think of something. Maybe add passing an instance of EnvConf to the Env-constructor.

njustesen:
> If you want to find out what the int means you can run env._compute_action(action_idx) which returns the corresponding Action object.

Cool. Can we have a function that does the opposite too? Action to int. That would allow algorithms to construct their own action space using domain knowledge.

> Make sense, I'll make it a game vs RandomBot as default and make a self-play switch in the constructor.

After thinking a bit more about it, in self-play you typically play against an earlier checkpoint and not the current policy (see https://njustesen.github.io/botbowl/a2c-selfplay.html) so maybe it's confusing to offer that kind of self-play. Different self-play methods use different ways of sampling and swapping these checkpoints, so maybe we can't have a default self-play implementation? What do you think?

mrbermell:
> > If you want to find out what the int means you can run env._compute_action(action_idx) which returns the corresponding Action object.
> 
> Cool. Can we have a function that does the opposite too? Action to int. That would allow algorithms to construct their own action space using domain knowledge.

I can for sure the reverse function too. I have something else in mind on this topic as well, I'll show it when it's done. 

> > Make sense, I'll make it a game vs RandomBot as default and make a self-play switch in the constructor.
> 
> After thinking a bit more about it, in self-play you typically play against an earlier checkpoint and not the current policy (see https://njustesen.github.io/botbowl/a2c-selfplay.html) so maybe it's confusing to offer that kind of self-play. Different self-play methods use different ways of sampling and swapping these checkpoints, so maybe we can't have a default self-play implementation? What do you think? 

The self play switch becomes "true self play", as you would want in a tree search or training on both sides. If you want to play against a checkpoint (i.e. Agent) then it should be possible set the opp_agent of the env (just as the previous env). I'll add this option. 

Though, I've not yet figured out how this should work in the Agent.act() function for a neural network based agent. I don't like how different training is from inference in the current a2c example. I'm considering different approaches for a remedy!

mrbermell:
I'll move the a2c_example's logging to TensorBoard. It's another dependency but I think it's pretty worthwhile.

mrbermell:
I'm approaching done with the a2c example using the new environment. I hope I've improved readability and made it more modular. 

There are still some small details to fix, mainly getting the self play working. And after that I'll re-run the training to make sure it replicates the results from previous. 

@njustesen can you make a rough review of the direction in a2c_example.py?

mrbermell:
I've made readability improvement and optimizations. a2c_examples.py runs 25% faster now for the smallest pitch size.  But... It's not learning, even worse it's de-learning, after a couple of optimizer steps the policy reliably makes fewer touchdowns than a bot that randomly picks actions from the action mask. I've been bashing my head at it for a while now, time for some sleep.

njustesen:
Try to observe every step in the 1v1 match and log rewards, network output, and the final actions. 
My guess is there could be a problem with the reward function (check both sides of the pitch) or the action translation.

mrbermell:
Finally found where the problem is. In CNNPolicy.evalate_actions the `F.log_softmax(policy, dim=1)` evaluates all '-inf' values to '-inf' instead of to zero, which of course creates NaNs futher down etc etc... I'll solve it soon. Thanks for the pointers!

mrbermell:
FINALLY!  Now it works as expected. I'm still going to polish on a few things and test running it for bigger sizes with self play enabled and scripted actions enabled.

mrbermell:
Status: 
1) I don't know why Github action failed the test. Very strange. I'll will look into it. 

2) The selfplay does not work. When executing A2CAgent.act(game) it freezes (without raising any errors) at 
```python 
class CNNPolicy: 
    def forward(spatial_input, non_spatial_input): 
        x1 = self.conv1(spatial_input) # doesn't get past this function call 
```
Only freezes in the worker though, in isolation it works just fine. I've been chasing any difference in the arguments or how the CNNPolicy is configured but nothing so far...

njustesen:
> I don't know why Github action failed the test. Very strange. I'll will look into it.

I think you forgot to add functools to setup.py.

njustesen:
I checked out this branch and ran the a2c_example.py with `selfplay = True`. I had to fix some issues that looks like this:

```
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 14, in <module>
    from a2c_agent import A2CAgent, CNNPolicy
  File "<fstring>", line 1
    (spatial_input.dtype=)
                        ^
SyntaxError: invalid syntax
```

But after that, it seems to work fine.
```
/Users/noju/opt/anaconda3/envs/botbowlii/bin/python /Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py
It works to initalize and use the self play agent before sending sending it to the workers


Save log to logs/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122.dat
Updates: 50, Episodes: 30, Timesteps: 2000, Win rate: 0.47, TD rate: 0.50, TD rate opp: 0.67, Mean reward: -0.448, Difficulty: 1.00
Save log to logs/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122.dat
Updates: 100, Episodes: 57, Timesteps: 4000, Win rate: 0.44, TD rate: 0.56, TD rate opp: 0.74, Mean reward: -0.145, Difficulty: 1.00
Save log to logs/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122.dat
Updates: 150, Episodes: 85, Timesteps: 6000, Win rate: 0.57, TD rate: 0.86, TD rate opp: 0.64, Mean reward: 0.278, Difficulty: 1.00
Save log to logs/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122.dat
Updates: 200, Episodes: 116, Timesteps: 8000, Win rate: 0.58, TD rate: 0.87, TD rate opp: 0.55, Mean reward: 0.596, Difficulty: 1.00
Saving models/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122_selfplay_1.nn
Swapping opponent to models/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122_selfplay_1.nn
Save log to logs/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122.dat
Updates: 250, Episodes: 143, Timesteps: 10000, Win rate: 0.57, TD rate: 1.07, TD rate opp: 0.85, Mean reward: 0.579, Difficulty: 1.00
Save log to logs/botbowl-1/8b52ab6e-7bd5-11ec-bbce-acde48001122.dat
Updates: 300, Episodes: 172, Timesteps: 12000, Win rate: 0.59, TD rate: 0.97, TD rate opp: 0.69, Mean reward: 0.252, Difficulty: 1.00
Process Process-1:
Process Process-2:
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 508, in <module>
    main()
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 318, in main
    spatial_obs, non_spatial_obs, action_masks, shaped_reward, tds_scored, tds_opp_scored, done = envs.step(action_objects, difficulty=difficulty)
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 200, in step
    results = [remote.recv() for remote in self.remotes]
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 200, in <listcomp>
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 159, in worker
    remote.send((spatial_obs, non_spatial_obs, action_mask, reward, tds_scored, tds_opp_scored, done))
      File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
results = [remote.recv() for remote in self.remotes]
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/connection.py", line 404, in _send_bytes
    self._send(header + buf)
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/connection.py", line 250, in recv
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/connection.py", line 368, in _send
    n = write(self._handle, buf)
KeyboardInterrupt
    buf = self._recv_bytes()
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/connection.py", line 407, in _recv_bytes
    buf = self._recv(4)
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/connection.py", line 379, in _recv
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/Users/noju/opt/anaconda3/envs/botbowlii/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py", line 138, in worker
    spatial_obs, reward, done, info = env.step(action)
  File "/Users/noju/git/tmp/botbowl/botbowl/ai/new_env.py", line 489, in step
    obs, reward, done, info = self.env.step(action, skip_observation)
    chunk = read(handle, remaining)  File "/Users/noju/git/tmp/botbowl/botbowl/ai/new_env.py", line 506, in step
    self.env.step(action, skip_observation=True)

  File "/Users/noju/git/tmp/botbowl/botbowl/ai/new_env.py", line 335, in step
    self.game.step(action)
  File "/Users/noju/git/tmp/botbowl/botbowl/core/game.py", line 181, in step
    self.action = self._safe_act()
KeyboardInterrupt  File "/Users/noju/git/tmp/botbowl/botbowl/core/game.py", line 319, in _safe_act
    action = self.actor.act(self)

  File "/Users/noju/git/tmp/botbowl/examples/a2c/a2c_agent.py", line 188, in act
    spatial_obs, non_spatial_obs, action_mask = tuple(map(transform_array, self.env.get_state()))
  File "/Users/noju/git/tmp/botbowl/botbowl/ai/new_env.py", line 228, in get_state
    index = count()
KeyboardInterrupt
libc++abi: terminating with uncaught exception of type std::__1::system_error: Broken pipe

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
```

Is this different from when you run it?

All I did was:
```
git checkout improved-env
conda create --name bbgym python=3.9
pip install -e .
```

Running on my macbook.

njustesen:
No, wait. Looks like I was still in the old anaconda environment. Interesting that that even worked.

In the correct environment, it still works, after installing torch too.

```
/Users/noju/opt/anaconda3/envs/bbgym/bin/python3.9 /Users/noju/git/tmp/botbowl/examples/a2c/a2c_example.py
It works to initalize and use the self play agent before sending sending it to the workers


Save log to logs/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122.dat
Updates: 50, Episodes: 25, Timesteps: 2000, Win rate: 0.48, TD rate: 1.52, TD rate opp: 1.28, Mean reward: -0.095, Difficulty: 1.00
Save log to logs/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122.dat
Updates: 100, Episodes: 50, Timesteps: 4000, Win rate: 0.56, TD rate: 1.08, TD rate opp: 1.04, Mean reward: -0.256, Difficulty: 1.00
Save log to logs/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122.dat
Updates: 150, Episodes: 75, Timesteps: 6000, Win rate: 0.72, TD rate: 1.76, TD rate opp: 0.80, Mean reward: 1.009, Difficulty: 1.00
Save log to logs/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122.dat
Updates: 200, Episodes: 99, Timesteps: 8000, Win rate: 0.77, TD rate: 1.67, TD rate opp: 0.71, Mean reward: 1.020, Difficulty: 1.00
Saving models/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122_selfplay_1.nn
Swapping opponent to models/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122_selfplay_1.nn
Save log to logs/botbowl-1/85ff3afa-7bd6-11ec-b02d-acde48001122.dat
Updates: 250, Episodes: 123, Timesteps: 10000, Win rate: 0.69, TD rate: 1.58, TD rate opp: 0.88, Mean reward: 0.444, Difficulty: 1.00
```

mrbermell:
That is an improvement from how it works on my setup. Can you push your commit to the branch? I'll check it out and take it from there.

Thanks so much for the help!

njustesen:
Done. Just fixed some comments so it will probably not make it run on your end. 

Here's my pip freeze dump:
```
(bbgym) ➜  botbowl git:(improved-env) ✗ pip freeze
attrs==21.4.0
-e git+https://github.com/njustesen/botbowl.git@9150978cd1e115e47a60316a9af8f473e484f692#egg=botbowl
certifi==2021.10.8
charset-normalizer==2.0.10
click==8.0.3
cloudpickle==2.0.0
cycler==0.11.0
Cython==3.0.0a10
Flask==2.0.2
fonttools==4.28.5
gym==0.21.0
idna==3.3
iniconfig==1.1.1
itsdangerous==2.0.1
Jinja2==3.0.3
kiwisolver==1.3.2
MarkupSafe==2.0.1
matplotlib==3.5.1
numpy==1.22.1
packaging==21.3
Pillow==9.0.0
pluggy==1.0.0
py==1.11.0
pyparsing==3.0.7
pytest==6.2.5
python-dateutil==2.8.2
python-interface==1.6.1
requests==2.27.1
six==1.16.0
stopit==1.1.2
toml==0.10.2
torch==1.10.1
typing_extensions==4.0.1
untangle==1.1.1
urllib3==1.26.8
Werkzeug==2.0.2
```

mrbermell:
What was a syntax error with my print statements is actually valid in python >= 3.8 but since we're supporting 3.6 I'll remove it before merging. 

> An f-string such as f'{expr=}' will expand to the text of the expression, an equal sign, then the representation of the evaluated expression. For example:

>>> user = 'eric_idle'
>>> member_since = date(1975, 7, 31)
>>> f'{user=} {member_since=}'
"user='eric_idle' member_since=datetime.date(1975, 7, 31)"

Anyway, I'll test some more. Thanks again!

njustesen:
Interestingly, the print syntax didn't work in my python 3.9 environment.

mrbermell:
> Interestingly, the print syntax didn't work in my python 3.9 environment.

Strange, it works for me. 
```
Python 3.9.7 (default, Sep 16 2021, 13:09:58) 
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = 123
>>> f"{a=}" 
'a=123'
```

mrbermell:
Finally got it! The fix was to initialize the `VecEnv` before initializing the `ac_agent`. They should be completely independent,.. I really wonder what happened. 

```python
#a2c_example.py

#This does not work 
def main(): 
    # .... 
    ac_agent = CNNPolicy(...) 
    envs = VecEnv([make_env() for _ in range(num_processes)])

#This works 
def main(): 
    # .... 
    envs = VecEnv([make_env() for _ in range(num_processes)])
    ac_agent = CNNPolicy(...) 

```

mrbermell:
As I was updating the a2c tutorial I had en epiphany. Instead of having the action_mask and non_spatial_obs in info we should actually make obs a tuple of all relevant state observations, it just makes more sense.  So like this (omitting the torch transformations for readability): 

```python 
env = BotBowlEnv() 
obs = env.reset() 
# spatial_obs, non_spatial_obs, action_mask = obs  # tuple unpacking if needed 
action_idx = ac_agent.act(*obs)
obs, reward, done, info = env.step(action_idx) 
```
I'll happily refactor into this, but I'd like your blessing first, @njustesen

ctelloordonez:
Hi, I'm getting an error when I enable the pathfinding in the EnvConf. If I enable it in the a2c_example.py script:

```
  File "C:\Users\Carlos\anaconda3\envs\botbowl\lib\multiprocessing\process.py", line 297, in _bootstrap
    self.run()
  File "C:\Users\Carlos\anaconda3\envs\botbowl\lib\multiprocessing\process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "c:\Users\Carlos\Documents\GitHub\botbowl\examples\a2c\a2c_example.py", line 137, in worker
    spatial_obs, reward, done, info = env.step(action)
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\env.py", line 499, in step
    obs, reward, done, info = self.env.step(action, skip_observation)
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\env.py", line 516, in step
    return self.root_env.get_step_return(skip_observation)
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\env.py", line 353, in get_step_return
    spatial_observation, non_spatial_observation, action_mask = self.get_state()
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\env.py", line 227, in get_state
    spatial_obs = np.stack([layer.get(game) for layer in self.env_conf.layers])
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\env.py", line 227, in <listcomp>
    spatial_obs = np.stack([layer.get(game) for layer in self.env_conf.layers])
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\layers.py", line 29, in get
    layer = self.produce(game)
  File "c:\users\carlos\documents\github\botbowl\botbowl\ai\layers.py", line 287, in produce
    roll = (action_choice.block_dice[i] + 3) / 6.0
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

Here are the action_choice and action_choice.block_dice prints before getting the error:
```
action_choice: ActionChoice(ActionType.MOVE, len(positions)=10, len(players)=0, len(paths)=10)
action_choice.block_dice: [None, None, None, None, None, None, None, None, None, None]
```

I couln't find where those None values are set so I'm not sure whether they should be all 0 in that case, if the list should be empty or if the actual values are missing there.

mrbermell:
thanks @ctelloordonez for testing this. I never did myself. I've pushed an update that solves it and a test. 

@njustesen , this fix adds small check in in botbowl.core.procedure.MoveAction._get_actions_from_paths(). I'm not very familiar with that. Can you have a look and confirm it makes sense?

njustesen:
As a final test since so much has changed, we should rerun all the experiments in the tutorials, compare and update the plots.

Are we ready for that now? I can run some of the longer runs on my computer.

mrbermell:
Yes, that would be good. I think it's ready for that. I haven't done the obs refactor yet, could wait for that but it's won't affect the overall logic and performance.

ctelloordonez:
There is something I've noticed when using this environment to replicate the mimicbot training. 

The `get_state()` method always flips the spatial_obs if needed, which also affects on the action_mask returned by it. Then also, `_compute_action(action_idx, flip)` could return the action with its position flipped if needed, while `_compute_action_idx(action)` will always return a non flipped position of the action.

I noticed this when, after collecting all the state-action pairs from both the home and away bot as the scripted one, and using the dataset to  train a network with Behavioral cloning, the resulting bot kept on returning illegal actions playing as the away team. That's because I set `flip=False` since I believed I was training my bot with information playing on both sides.

In that situation the problem of the illegal actions could be fixed with `_compute_action(action_idx, flip=env._flip_x_axis())`. But I still had the problem of having stored all my action_indexes with `_compute_action_idx(action)` which as I've mention never flips the position of the action before computing its index.

I know it's a bit confusing the way I've described it, but I think a good solution in this case is to add to both, `get_state()` and `_compute_action_idx(action)` a `flip` parameter as done in `_compute_action(action_idx, flip)` and let the user decide whether they want to flip the positions when possible, or always work with a fixed position distribution.

mrbermell:
@ctelloordonez , have a look at the change log in this [commit](https://github.com/njustesen/botbowl/pull/197/commits/1669af1aaa0c82804d2ca693aadabbffe5bd0fa5) 

To summarize, 

1. I renamed `_flip_x_axis()` to `away_team_active()`. It's a more descriptive name. I also added `home_team_active()` for symmetry. 
2. I've added optional argument `flip: Optional[bool]=None`, to `get_state()`, `_compute_action()` and `_compute_action_idx()` as you suggested. 
In the start of each function we have `flip=self.away_team_active() if flip is None else flip`. This means, if you never provide the flip argument, the environment will always behave as though you're playing as the home team. 

Finally, I made the agreed change of moving the non spatial observation and action mask into the observation. I consider renaming `get_state()` to `get_observation()`, it just makes more sense. 

Let me know what you think.

njustesen:
Are we just missing torch from setup.py or is there another reason why it's not in it?

```
  File "/home/niels/git/botbowl/examples/a2c/a2c_example.py", line 7, in <module>
    import torch
ModuleNotFoundError: No module named 'torch'
```

mrbermell:
I didn't add it torch to the list of needed requirements because it would weigh down the framework for everyone. I think we discussed it at some point. I'm not against adding it.

njustesen:
I vaguely that now. Let's just take sure it's written in the top of a2c_example.py. 

Btw. I am re-running all the RL experiments in the tutorials. So far it's basically the same as before.

njustesen (commented):
This looks oddly specific to the 11 player variant. Does it also work for other variants?

botbowl/core/pathfinding/cython_pathfinding.pyx
```diff
@@ -152,6 +152,13 @@ cdef class Pathfinder:
     cdef priority_queue[NodePtr] open_set
     cdef mapcpp[double, vector[NodePtr]] risky_sets 
 
+    def __dealloc__(self):
+        cdef NodePtr empty_ptr
+        for i in range(17):
+            for j in range(28):
+                self.locked_nodes[i][j] = empty_ptr
+                self.nodes[i][j] = empty_ptr
+
```
njustesen, line 161: This looks oddly specific to the 11 player variant. Does it also work for other variants?

njustesen (commented):
This looks oddly specific to the 11 player variant. Does it also work for other variants?

botbowl/core/pathfinding/cython_pathfinding.pyx
mrbermell: It's fine. `locked_nodes` and `nodes` have static size. I choose the size according to the biggest pitch size. If you use a smaller pitch size the out of bounds nodes will just sit there and never be used. I thought this was a better solution than nestling cpp vectors to get a dynamic sizing. 

```python
cdef class Pathfinder:
    .... 
    cdef NodePtr locked_nodes[17][28] # initalized as empty pointers
    cdef NodePtr nodes[17][28] # initalized as empty pointers
```
 
The C equivalent is 
```C
struct {
    ....
    NodePtr    locked_nodes[17][28]; 
    NodePtr    nodes[17][28]; 
} Pathfinder; 
``` 

The cause of the memory leak was that NodePtr's destructor was not called for the nodes in `locked_nodes` and `nodes`, because their type is `NodePtr **` and they don't even know the size of what they're pointing to. And because NodePtr is a `std::shared_ptr` it has the responsibility to free memory on the heap, which was not done. My solution makes sure that the destructor is called.

njustesen: Okay. That's fine for now. Just keep in mind we should be able to support dungeon bowl at some point :)

examples/gym_example.py
```diff
@@ -11,16 +11,14 @@ def main():
...
-        _, _, mask = env.get_state()
+        _, _, mask = env.reset()
 
         while not done:
             env.render(feature_layers=True)
             aa = np.where(mask > 0.0)[0]
             action_idx = np.random.choice(aa, 1)[0]
-            _, _, done, info = env.step(action_idx)
-            mask = info['action_mask']
+            (_, _, mask), _, done, _ = env.step(action_idx)
```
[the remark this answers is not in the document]
mrbermell: They're unnamed because they're not used. But since this is an example, names should be explicit. I'll update

botbowl/ai/new_env.py
```diff
@@ -22,13 +22,21 @@ def take(n: int, iterable: Iterable) -> None:
 
 
 class EnvConf:
```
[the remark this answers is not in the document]
mrbermell: This comment is label "Outdated" and attached to a file that doesn't exist anymore. This confuses me. 

In the beginning EnvConf was just a singleton but some weeks ago I changed it to an instantiable class with a constructor filled with interesting arguments. Look in the latest files and you'll see :)

njustesen:
I get this when I run with self play and pathfinding. 

```
Process Process-5:
  File "/home/niels/anaconda3/envs/botbowl/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/niels/anaconda3/envs/botbowl/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/niels/git/botbowl/examples/a2c/a2c_example.py", line 137, in worker
    (spatial_obs, non_spatial_obs, action_mask), reward, done, info = env.step(action)
  File "/home/niels/git/botbowl/botbowl/ai/env.py", line 539, in step
    obs, reward, done, info = self.env.step(action, skip_observation)
  File "/home/niels/git/botbowl/botbowl/ai/env.py", line 554, in step
    self.env.step(action, skip_observation=True)
  File "/home/niels/git/botbowl/botbowl/ai/env.py", line 357, in step
    self.game.step(action)
  File "/home/niels/git/botbowl/botbowl/core/game.py", line 191, in step
    self.action = self._safe_act()
  File "/home/niels/git/botbowl/botbowl/core/game.py", line 328, in _safe_act
    action = self.actor.act(self)
  File "/home/niels/git/botbowl/examples/a2c/a2c_agent.py", line 173, in act
    self._filter_actions()
  File "/home/niels/git/botbowl/examples/a2c/a2c_agent.py", line 146, in _filter_actions
    roll = action_choice.paths[i].rolls[0]
IndexError: list index out of range
```

mrbermell:
> I get this when I run with self play and pathfinding.

Hmm.. _filter_actions() shouldn't even be called if pathfinding is enabled. I'll look into it!

mrbermell:
I've pushed a solution. When I tried to update _filter_actions() it struck me as odd. Does it achieve do what it sets out to do, namely filter non adjacent move actions? I think the should be a distance check somewhere. 

```python 
def _filter_actions(self):
    """
    Remove pathfinding-assisted non-adjacent or block move actions if pathfinding is disabled.
    """
    actions = []
    for action_choice in self.env.game.state.available_actions:
        if action_choice.action_type == ActionType.MOVE:
            positions, block_dice, rolls = [], [], []
            for i in range(len(action_choice.positions)):
                position = action_choice.positions[i]
                roll = action_choice.paths[i].rolls[0]
                # Only include positions where there are not players
                if self.env.game.get_player_at(position) is None:
                    positions.append(position)
                    rolls.append(roll)
            actions.append(ActionChoice(ActionType.MOVE, team=action_choice.team, positions=positions, rolls=rolls))
        else:
            actions.append(action_choice)
    self.env.game.state.available_actions = actions
```

examples/a2c/a2c_example.py
```diff
@@ -18,10 +18,11 @@
 # Environment
 env_size = 3  # Options are 1,3,5,7,11
 env_name = f"botbowl-{env_size}"
-env_conf = EnvConf(size=env_size, pathfinding=False)
+env_conf = EnvConf(size=env_size, pathfinding=True)
 
 
-make_agent_from_model = partial(A2CAgent, env_conf=env_conf, scripted_func=a2c_scripted_actions)
+make_agent_from_model = partial(A2CAgent, env_conf=env_conf, scripted_func=a2c_scripted_actions,
+                                exclude_pathfinding_moves=False)
```
njustesen, line 25: pathfinding should be False by default, I think. Maybe we should set `exclude_pathfinding_moves= not env_conf.pathfinding`

examples/a2c/a2c_example.py
```diff
@@ -18,10 +18,11 @@
 # Environment
 env_size = 3  # Options are 1,3,5,7,11
```
njustesen, line 19: Let's set this to 1 by default too.

njustesen (changes requested):
Just minor things to make sure it in line with the first tutorial.

njustesen:
I ran all the experiments from the tutorials and they are identical so I don't think we even need to update them.

njustesen:
> I've pushed a solution. When I tried to update _filter_actions() it struck me as odd. Does it achieve do what it sets out to do, namely filter non adjacent move actions? I think the should be a distance check somewhere.
> 
> ```python
> def _filter_actions(self):
>     """
>     Remove pathfinding-assisted non-adjacent or block move actions if pathfinding is disabled.
>     """
>     actions = []
>     for action_choice in self.env.game.state.available_actions:
>         if action_choice.action_type == ActionType.MOVE:
>             positions, block_dice, rolls = [], [], []
>             for i in range(len(action_choice.positions)):
>                 position = action_choice.positions[i]
>                 roll = action_choice.paths[i].rolls[0]
>                 # Only include positions where there are not players
>                 if self.env.game.get_player_at(position) is None:
>                     positions.append(position)
>                     rolls.append(roll)
>             actions.append(ActionChoice(ActionType.MOVE, team=action_choice.team, positions=positions, rolls=rolls))
>         else:
>             actions.append(action_choice)
>     self.env.game.state.available_actions = actions
> ```

This does look weird. I will take a look.

njustesen:
I think `_filter_actions()` is from when we had the MOVE action type for handoffs, blitzes, and stuff if the action included a path, but we forgot to remove it in https://github.com/njustesen/botbowl/pull/174 since it's no longer needed. 

Looks like everything related to self-play and pathfinding is working fine without the function. 

However, we need to update a2c_agent.py and the tutorial https://njustesen.github.io/botbowl/a2c-full.html. It doesn't run out of the box.

examples/a2c/a2c_example.py
mrbermell: done.

examples/a2c/a2c_example.py
mrbermell: done

examples/gym_example.py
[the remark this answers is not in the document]
mrbermell: done

botbowl/core/procedure.py
```diff
@@ -2710,7 +2710,7 @@ def _get_actions_from_paths(self, paths):
         for action_type, action_paths in paths_by_type.items():
             if len(action_paths) > 0:
                 positions = [path.get_last_step() for path in action_paths]
-                block_dice = [path.block_dice for path in action_paths]
+                block_dice = [path.block_dice for path in action_paths if path.block_dice is not None]
```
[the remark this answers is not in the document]
mrbermell: done

mrbermell:
My interpretation is that `_filter_actions()` is not intended to be used during training, only inference. I guess you still need it for you bot in the comp if you trained your RL agent without pf?  

> However, we need to update a2c_agent.py and the tutorial https://njustesen.github.io/botbowl/a2c-full.html. It doesn't run out of the box.

What needs updating? 

By the way, this branch is in the main repo so you can push to it if you want. I don't mind receiving review comments and fixing according to them but you're more welcome to edit details directly :)

examples/a2c/a2c_example.py
njustesen: We should be able to remove `exclude_pathfinding_moves` completely now that we found _filter_actions() to be useless, right?

mrbermell:
I've updated the multi_gym_renderer.py. It's rendering in different windows.

njustesen:
I can't figure out why I get this error when running a2c_agent.py. How come a2c_example.py with `selfplay=True` is able to load an agent from a model file but this script is not?

```
/Users/noju/opt/anaconda3/envs/botbowl/bin/python /Users/noju/git/botbowl/examples/a2c/a2c_agent.py
Starting game 1
  File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 232, in <module>
    main()
  File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 212, in main
    game.init()
  File "/Users/noju/git/botbowl/botbowl/core/game.py", line 151, in init
    self.step(start_action)
  File "/Users/noju/git/botbowl/botbowl/core/game.py", line 191, in step
    self.action = self._safe_act()
  File "/Users/noju/git/botbowl/botbowl/core/game.py", line 328, in _safe_act
    action = self.actor.act(self)
  File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 157, in act
    _, actions = self.policy.act(
  File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 90, in act
    values, action_probs = self.get_action_probs(spatial_inputs, non_spatial_input, action_mask=action_mask)
  File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 106, in get_action_probs
    values, actions = self(spatial_input, non_spatial_input)
  File "/Users/noju/opt/anaconda3/envs/botbowl/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 77, in forward
    x3 = self.linear1(concatenated)
  File "/Users/noju/opt/anaconda3/envs/botbowl/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
  File "/Users/noju/opt/anaconda3/envs/botbowl/lib/python3.9/site-packages/torch/nn/modules/linear.py", line 103, in forward
    return F.linear(input, self.weight, self.bias)
  File "/Users/noju/opt/anaconda3/envs/botbowl/lib/python3.9/site-packages/torch/nn/functional.py", line 1848, in linear
    return torch._C._nn.linear(input, weight, bias)
RuntimeError: mat1 and mat2 shapes cannot be multiplied (1x30592 and 2048x128)
```

Could you take a look @mrbermell ?

mrbermell:
TL DR:  You're loading a model that was trained on 1v1-board. But you're trying to make it play on 11v11-board. 

Investigation notes: 

> ```
>   File "/Users/noju/git/botbowl/examples/a2c/a2c_agent.py", line 77, in forward
>     x3 = self.linear1(concatenated)
>     .....
> RuntimeError: mat1 and mat2 shapes cannot be multiplied (1x30592 and 2048x128)
> ```

I interpret this as `concatenated` has the size1x30592 and `linear1` has size 2048x128.  Since linear1 is defined as: 

```python
#a2c_example.py
num_hidden_nodes = 128
num_cnn_kernels = [32, 64]

# a2c_agent.py 
class CNNPolicy(nn.Module): 
    def __init__(...): 
        .... 
        stream_size = kernels[1] * spatial_shape[1] * spatial_shape[2]
        stream_size += hidden_nodes
        self.linear1 = nn.Linear(stream_size, hidden_nodes)
```

So I would expect streamsize to be 64x17x26+128 = 30592 in 11v11.  If you instead plug in the numbers for 1v1 you get stream size = 64x6x5+128 = 2048. So you've trained on 1v1 and loaded that model and tried to play in on 11v11.

njustesen:
That would explain the error. I am loading a model trained on 1v1 but I am also setting `env_size = 1`. Not sure how I can make sure the model is for 1v1, if that's not enough.

mrbermell:
Maybe I'm too tired to get this. But you have `env_size=1` in a2c_agent.py, so you're gonna load 1v1-model. But the main() function still creates a 11v11 game. 

```python
def main():
    # Register the bot to the framework
    def _make_my_a2c_bot(name, env_size=env_size):
        return A2CAgent(name=name,
                        env_conf=EnvConf(size=env_size),
                        scripted_func=a2c_scripted_actions,
                        filename=model_filename)
    botbowl.register_bot('my-a2c-bot', _make_my_a2c_bot)

    # Load configurations, rules, arena and teams
    config = botbowl.load_config("bot-bowl")
```
That last row results in a 11v11-game, right?

njustesen:
Thanks! Turns out I was the tired one. It's working.

I only have one thing left on this PR before I think we can merge. I noticed the a2c_scripted_actions which is added to the a2c_example.py. It's a great feature but maybe it shouldn't be enabled by default? Perhaps we could disable it and write a short tutorial on how to apply scripted actions to RL. What do you think @mrbermell ?

mrbermell:
Great. I only added scripted behavior that is almost always correct so it could be on by default. But I don't have a strong opinion, you may choose :)

PR #198, opened 2022-01-02
Python 3.10 added to github workflow

### LINKED ISSUE #173: Add support for python 3.9 and 3.10
Check if tests pass with python 3.9 and 3.10 and add these versions to the workflow script.

njustesen:
Fixes #173.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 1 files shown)


mrbermell (approved):
Looks good :) 
I took the liberty of change the python version to '3.10', it's parsed as a number so 3.10 became 3.1

EDIT:
I also found a good argument to keep python 3.6.  Some old docker images might still use it. If it turned out my SEED RL Docker image uses 3.6 I don't think it would necessarily be a smooth ride to patch it.

PR #202, opened 2022-01-17
Fixes bugs in MyScriptedBot

mrbermell:
Fixes bugs raised by user "carlosto" on Discord (@ctelloordonez on github). 

1. Introduces function `path_to_move_actions(game, path) -> List[Action]` which converts a Path returned from the pathfinding algorithms to actions, including blitz, handoff and foul. It makes the code more readable and makes the bot agnostic to whether the current game has pathfinding enabled or not. 

2. Solves bug with None being returned by `MyScriptedBot.setup()`. Solution found by _carlosto_. 

3. Introduces test for scripted bot, the simply runs a game between two scripted bots.

### DIFF
--- a/botbowl/ai/proc_bot.py
+++ b/botbowl/ai/proc_bot.py
@@ -5,7 +5,8 @@ Year: 2018
 ==========================
 This module contains an example bot that takes random actions.
 """
-from botbowl.core.model import Agent
+from botbowl.core.model import Agent, Action
+from botbowl.core.game import Game, InvalidActionError
 from botbowl.core.procedure import *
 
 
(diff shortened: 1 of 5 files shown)


njustesen (changes requested):
Besides the test.can you also check that the win rate is similar to what is reported in the tutorial?

Can you write in the PR what the changes does?

In the first comment in the game class, please use "agent" instead of "bot". Its unclear what bot is.

Otherwise, I think it looks good.

ctelloordonez:
I have run some games with MyScriptedBot on this PR and it still returns not allowed actions. An example of this is the following error with a list of available actions on that state

```
Available actions: [ActionChoice(ActionType.MOVE, len(positions)=97, len(players)=0, len(paths)=97), ActionChoice(ActionType.STAND_UP, len(positions)=0, len(players)=0, len(paths)=0), ActionChoice(ActionType.END_PLAYER_TURN, len(positions)=0, len(players)=0, len(paths)=0)]

  File "c:/Users/Carlos/Documents/GitHub/botbowl/examples/scripted_bot_example.py", line 752, in <module>
    game.init()
  File "c:\users\carlos\documents\github\botbowl\botbowl\core\game.py", line 138, in init
    self.step(start_action)
  File "c:\users\carlos\documents\github\botbowl\botbowl\core\game.py", line 158, in step
    done = self._one_step(self.action)
  File "c:\users\carlos\documents\github\botbowl\botbowl\core\game.py", line 396, in _one_step
    f"Action not allowed {action.to_json() if action is not None else 'None'}")
botbowl.core.game.InvalidActionError: Action not allowed {'action_type': 'MOVE', 'position': {'x': 14, 'y': 11}, 'player_id': None}
```

njustesen:
I believe the scripted bot has always returned illegal actions once in a while. It was just never an issue before we made the game class throw an exception. 

These bugs can be pretty hard to track down. Hopefully, we can find it but another option is to check if every action is legal and if not, end the player turn, or the entire turn.

ctelloordonez:
I have been working on the 0.3.1 release branch the past months and it did return some illegal actions. 

What I've done to overcome the problem while its being fixed is instead of throwing an exception on game.py when trying to do an illegal action is just pick another random action from the available ones. 

This doesn't seem to affect much the performance of the scripted bot. Scores an average of 4-0 TD against the random bot.

Of course the "hotfix" I mentioned on game.py shouldn't be an option but maybe scripted bot could check if it is going to return an illegal action. In that case just return another random one, or as you've suggested end the player or the entire turn.

mrbermell:
I think it's an improvement to have a stricter game engine that crashes. It should be the responsibility of the agent to safe guard against illegal action and handle them accordingly. The agent should check game._is_action_allowed() but we should make that a public function. 

However, the scripted bot we provide in the framework should never suggest illegal actions, but we'll implement the safe guard so other can use it. I'll see what I can do! 

I didn't run into any troubles when I was testing yesterday but I can't claim that I was rigorous. Regarding the error you showed. You are running the scripted bot without your own modifications as it is looking the the PR's branch, correct?  Is it only MOVE-actions that are reported as illegal or is it other actions too?  Could be the pathfinding that is doing something strange...

njustesen:
Maybe it should be optional if Game throws an exception?
The problem is that the bot developers cannot catch the error and it's difficult to develop a 100% correct bot.

At some point, I would like to flip around the relationship between Game and Agents. The Game should not query the Agents for actions. Instead, a third entity (e.g. a Competition class, the GymEnv, or a custom script) should be responsible for querying the Agents, stepping in the Game, and handling any exceptions that are thrown. This is a big change though, so maybe we can consider it for next year. It would, however, resolve this issue because you can catch the exception.

mrbermell:
If I rewrite the proc_bot or scripted's act() function to always check game._is_action_allowed(), it will be easy to identify and also handle. Having a switch in the game class to "opt out" of the exception throwing is fine too but personally I wouldn't use it, I like the strictness too much 🤓

ctelloordonez:
_> I didn't run into any troubles when I was testing yesterday but I can't claim that I was rigorous. Regarding the error you showed. You are running the scripted bot without your own modifications as it is looking the the PR's branch, correct? Is it only MOVE-actions that are reported as illegal or is it other actions too? Could be the pathfinding that is doing something strange..._

Yes I'm running the scripted bot from the PR's branch without modifications. It did reported BLOCK-actions as illegal also.

I also agree that it is better to have an stricter game class, so that it is on the bots' end not to return an illegal action.

njustesen:
Okay, I think your suggestion is good for now, until we have a better structure.

mrbermell:
Also, I ran the statistics as you requested.  The scripted bot wins all games against random with approx 4.0 touchdowns per game now. The tutorial says 3.5, I think it's a small improvement that I and my strictness take all the credit for! :smile:

njustesen (changes requested):
Cool. Add a line or two in the tutorial that describes the change and update the code examples in the tutorial. Then I guess it's done?

njustesen (approved):

PR #196, opened 2021-12-21
[WiP] Fixes take root stand up bug

njustesen:
Fixes #187 by allowing take root players to stand up.

This required some small changes to the pathfinding code but my python changes are not reflected in the cython pathfinder when I build. Can you please help @mrbermell ?

The two tests that are failing are:
- tests.ai.test_pathfinding.test_compare_cython_python_paths
- tests.framework.test_forward_model.test_deepcopy_of_game_is_correct

Issue #187, opened 2021-11-18
Take Root shall not prevent standing up


mrbermell:
A prone player with skill TAKE_ROOT shall be able to stand up even if Take Root roll failed. Below test should pass but currently does not. 

``` python 
def test_taken_root_players_can_stand_up():
    game = get_game_turn()
    team = game.get_agent_team(game.actor)
    team.state.rerolls = 0  # ensure no reroll prompt

    players = game.get_players_on_pitch(team)
    player = players[1]
    player.extra_skills = [Skill.TAKE_ROOT]
    player.state.up = False

    D6.FixedRolls.clear()
    D6.fix(1)  # fail take root
    D6.fix(4)  # succeed stand up roll

    game.step(Action(ActionType.START_MOVE, player=player))
    game.step(Action(ActionType.STAND_UP, player=player))

    assert player.state.up
    assert player.state.taken_root
```

njustesen:
The problem is that when taking root, the player turn ends. This test passes but it's not correct.

```python
def test_take_root_ends_move_turn():
    game = get_game_turn()
    team = game.get_agent_team(game.actor)
    team.state.rerolls = 0  # ensure no reroll prompt

    players = game.get_players_on_pitch(team)
    player = players[1]
    player.extra_skills = [Skill.TAKE_ROOT]

    D6.FixedRolls.clear()
    D6.fix(1)  # fail trait test

    game.step(Action(ActionType.START_MOVE, player=player))

    # check the player turn has not ended
    assert game.state.active_player is not player
```

PR #204, opened 2022-01-24
Fixes take root stand up bug (2) 

mrbermell:
Solves issue #187 and replaces PR #196

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
 
     steps:
     - uses: actions/checkout@v1
--- a/README.md
+++ b/README.md
@@ -16,9 +16,6 @@ Please cite our paper if you use botbowl in your publications.
 
 Read more about the challenges and achievements in [Papers.md](docs/papers.md)
 
-## Bot Bowl - the compitition
-Bot Bowl is an AI competition using the Bot Bowl framework. Go read all about the results of [Bot Bowl III](docs/bot-bowl-iii.md). Not to be confused with 'botbowl' which refers to the module/repo. 
-
 ## Installation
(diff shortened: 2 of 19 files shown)


mrbermell:
I think this is ready to merge and quite important because is has crucial fixes for the scripted bot too.

mrbermell:
There's a problem with this the additions in the Player-class. num_moves_left() can return negative values if the `include_gfi`is not set. I think that is not how you would expect the function to work. It also brakes the "movement left" layer which can then return negative values.  And num_gfi_left() seems to work according to the principle that num_moves_left returns negative values. 

```python
def num_moves_left(self, include_gfi: bool = False):
    if self.state.taken_root:
        return 0
    if self.state.used or self.state.stunned:
        moves = 0
    else:
        moves = self.get_ma()
        moves = moves - self.state.moves
        if include_gfi:
            if self.has_skill(Skill.SPRINT):
                moves = moves + 3
            else:
                moves = moves + 2
    return max(0, moves)

def num_gfis_left(self):
    ma = self.num_moves_left()
    gfis_used = 0 if ma >= 0 else -ma
    if self.has_skill(Skill.SPRINT):
        return 3 - gfis_used
    elif self.state.taken_root:
        return 0
    return 2 - gfis_used
```

mrbermell:
I've pushed my solution proposal: 

```python
def num_moves_left(self, include_gfi: bool = False):
    if self.state.taken_root or self.state.used or self.state.stunned:
        return 0
    moves = self.get_ma() - self.state.moves
    if include_gfi:
        if self.has_skill(Skill.SPRINT):
            moves = moves + 3
        else:
            moves = moves + 2
    return moves

def num_gfis_left(self):
    return self.num_moves_left(include_gfi=True) - self.num_moves_left(include_gfi=False)
```

mrbermell:
> Looks right but check if anything is expecting it to be lease than 0 if GFI is used. This would be bad of course but I have a vague memory of doing this at some point.

Only `num_gfis_left()` which used the negative values, but as stated, that is fixed. I did check the rest of the code, it's fine.

njustesen:
There are no more teams with a Treeman?

mrbermell:
Huh?

njustesen (approved):
Looks good!

Screenshot 2022-02-14 at 00 02 47

njustesen:
> Huh?

Nevermind :)

PR #205, opened 2022-01-30
Saving, loading, and deleting works

### LINKED ISSUE #87: Saving a game doesn't work

njustesen:
Fixes #87 by changing small things in existing code.

Screenshot 2022-01-30 at 15 24 26

Some UX improvements could be made but I don't want to spend further time on it now, e.g.:
- Angular-like page updating to avoid flickering
- The table rows and buttons don't align well 
- Right now, the web page doesn't jump into the game automatically when it's loaded to allow the user to select team/hotseat but the timer starts imediately. This could be impemented when we have a pause/resume feature.

### DIFF
--- a/botbowl/core/game.py
+++ b/botbowl/core/game.py
@@ -524,6 +524,14 @@ class Game:
             if clock.is_running():
                 clock.pause()
 
+    def resume_clocks(self) -> None:
+        """
+        Resume all clocks.
+        """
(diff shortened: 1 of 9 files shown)


mrbermell (approved):
I tested saving a loading and saving games in different stages. Good new feature!

PR #206, opened 2022-02-12
Fix blitz with one move left

mrbermell:
I've corrected a bug in the BlitzAction-procedure. It  wrongly set the gfi flag when blitzing with one move left, see below test which currently fails in master.  

This PR is built on #204, So that PR should be merged first.  

```python 
@pytest.mark.parametrize("pf_enabled", [False, True])
def test_blitz_one_move_left(pf_enabled):
    game, (player, opp_player) = get_custom_game_turn(player_positions=[(5, 5)],
                                                      opp_player_positions=[(6, 6)],
                                                      pathfinding_enabled=pf_enabled)

    player.role.ma = 1

    game.step(Action(ActionType.START_BLITZ, player=player))
    assert player.num_moves_left() == 1

    game.step(Action(ActionType.BLOCK, position=opp_player.position))

    assert not game.has_report_of_type(botbowl.OutcomeType.FAILED_GFI)
    assert not game.has_report_of_type(botbowl.OutcomeType.SUCCESSFUL_GFI)
```

PS. How did I find this bug? My deterministic tree searcher is very coupled with procedure.py and the pathfinding implementation (enough to make uncle Bob cry...) which means that it crashes when there are discrepancies between the two.


njustesen (approved):
Not sure how this bug could exist in the code but great you caught it!

Remember to pull this only when the test is passing, which is when you have merged the other PR, I suppose.

Issue #208, opened 2022-02-21
pathfinding shows path that continue after touchdown


mrbermell:
The pathfinding algorithm doesn't check for touchdown and keeps expanding. This leads to MOVE-actions that have paths that will never be completed. 

This is usually not a problem because the TD-proc is still handled. But it shouldn't be there and it's a simple fix. I'll do it soon.

njustesen:
Nice catch. It should speed up the algorithm a bit too.

mrbermell:
issues solved and merged to master.

PR #209, opened 2022-02-24
Pathfinding fixes - bugs #208, #211

mrbermell:
solve #208:

> The pathfinding algorithm doesn't check for touchdown and keeps expanding. This leads to MOVE-actions that have paths that will never be completed.

and solve #211 

> During a MoveAction, I think you should only pickup the ball if you select the square with the ball in. i.e. the pathfinding should stop expanding its search if it picks up the ball. It currently doesn't.

### DIFF
--- a/botbowl/core/pathfinding/python_pathfinding.py
+++ b/botbowl/core/pathfinding/python_pathfinding.py
@@ -193,6 +193,8 @@ class Pathfinder:
         self.can_block = can_block
         self.can_handoff = can_handoff
         self.can_foul = can_foul
+        self.carries_ball = None
+        self.endzone_x = None
         self.ma = player.num_moves_left()
         self.gfis = player.num_gfis_left()
         self.locked_nodes = np.full((game.arena.height, game.arena.width), None)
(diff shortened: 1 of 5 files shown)


mrbermell:
@njustesen , please review and approve :)

botbowl/core/pathfinding/python_pathfinding.py
```diff
@@ -309,6 +315,9 @@ def _expand(self, node: Node, target=None):
         if self.carries_ball and node.position.x == self.endzone_x:
             return
 
+        if (not self.carries_ball) and node.position == self.game.get_ball_position():
```
njustesen, line 318: The ball can be in the air, e.g. during Blitz or Quick Snap! We should make a check for that.

njustesen (changes requested):
Looks good. Just a small thing regarding balls in the air.

If you make a quick smoke test in the web app and the pathfinding works well there, I think it's good to go after the fix.

botbowl/core/pathfinding/python_pathfinding.py
mrbermell: Done.

mrbermell:
Solved your review comment and smoke test in web app is is fine after playing a drive against myself.

njustesen (approved):

PR #212, opened 2022-03-08
Bugfix pathfinding

mrbermell:
I've found some typos in the cython pathfinding. Since I'm poking around in the pathfinding I will also: 

- [x] Write tests that fail without this fix, and more pf tests in general
- [x] Refactor cython to use `make_shared<Node>(...)` instead of  `NodePtr(new Node(...))`
- [x] Implement "Snow for it" in cython and python pathfinding (GFI should be 3+ instead of 2+ if weather is blizzard)
- [x] MoveAction-procedure not modify the path object, should only modify a copy of `path.steps`. Consider making `path.steps` a tuple which is an immutable container.

### DIFF
--- a/botbowl/core/pathfinding/cython_pathfinding.pyx
+++ b/botbowl/core/pathfinding/cython_pathfinding.pyx
@@ -66,7 +66,7 @@ cdef class Path:
         public double prob
         public object block_dice, handoff_roll, foul_roll
 
-    def __cinit__(NodePtr n):
+    def __cinit__(self, NodePtr n):
         self.final_node = n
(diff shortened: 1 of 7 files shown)


njustesen:
Are any of these issues critical? Was planning to make a release now for the announcement.

mrbermell:
> Are any of these issues critical? Was planning to make a release now for the announcement.

Nope. Nothing a regular user would notice. The probabilities in the path are calculated correctly but the rolls sometimes in the wrong place, that's it.

njustesen (approved):
Looks good to me.

PR #214, opened 2022-03-10
Fix to forced setup

### LINKED ISSUE #213: Forced action does not work during setup
Forced action when timing out doesn't work when you haven't fielded any players. It's just stuck here:
Screenshot 2022-03-10 at 10 45 52

njustesen:
Fixes #213 by forcing random actions (except PLACE_PLAYER) until setup is legal.


mrbermell (approved):
Code checked. I tested in web UI and waited for some timeouts and it seems to work as expected. Well done. Ready to merge in my opinion :)

PR #217, opened 2022-03-14
Smaller variants working in web app

### LINKED ISSUE #216: Add support for all variants in web app

### LINKED ISSUE #221: Missing undo action with all actions but Move Action
This is after starting a blitz action. No undo action.

Screenshot 2022-03-29 at 21 39 06

njustesen:
Fixes #216 by adding support for all the smaller board sizes. Lot's of ugly styling and html but actually deleted some javascript as well.

Also fixes #221 by calling the super class start() function. This is here can_undo is set.

After playing a bit, I always get touchback so maybe the 3D scatter dice is not working.

Screenshot 2022-03-14 at 23 36 38
Screenshot 2022-03-14 at 23 37 07
Screenshot 2022-03-14 at 23 37 36
Screenshot 2022-03-14 at 23 39 44
Screenshot 2022-03-14 at 23 40 30


mrbermell:
Really nice to be able to play and watch bots play in the smaller variants. Here are my remarks (enumerate for convenience): 

1) D3 scatter dice is not working, It's a typo on row 3391 in procedure.py (see code below). But I think even D3 is too much for 1v1 and possibly 3v3. For 1v1 is should just be 1 scatter distance, we'll still see many touchbacks. For 3v3 it could be smallest of two D3 rolls. 
```python 
if self.game.config.kick_scatter_dice == 'd6':
    distance_dice = D6(self.game.rnd)
elif self.game.config.kick_scatter_dice == 'd3':
    distance_dice = D6(self.game.rnd)  # should be D3 here. 
```

2) Could consider limiting the throw in distance dice for smaller boards too. My proposal would be: 1v1: just 1.  3v3: D3.  5v5: D6.   Others unchanged. 

3) Scripted bot should work on smaller pitch sizes. I think it's just the setup actions and PLACE_BALL action that needs fixing. 

What do you think?

njustesen:
Thanks. Great list. I'll look into it!

mrbermell:
I'm not sure it's a good idea to keep an rng in the Die-classes. I think we should limit the game engine's rng use the the rng located in Game class. That is the assumption of `tests.util.only_fixed_rolls()` that was introduced some months ago.

njustesen:
1-4 done now. 

Not sure what you mean with the rng? [The dice models](https://github.com/njustesen/botbowl/blob/1717148055261776407f1c00e4a1d06ed81ffafb/botbowl/core/model.py#L806) already had this before:

mrbermell:
> 1-4 done now.
> 
> Not sure what you mean with the rng? [The dice models](https://github.com/njustesen/botbowl/blob/1717148055261776407f1c00e4a1d06ed81ffafb/botbowl/core/model.py#L806) already had this before:

The newly added constructor for `Die` and modified constructor of `D6`: 

```python
class Die(ABC):
    def __init__(self, rng):
        self.rng = rng if rng is not None else np.random.RandomState()

class D6(Die, Immutable):
    def __init__(self, rng=None):
        super().__init__(rng)
        if len(D6.FixedRolls) > 0:
            self.value = D6.FixedRolls.pop(0)
        else:
            self.value = self.rng.randint(1, 7)

```
makes it so that a call to `D6()` will initialize `self.rng` to be a completely new rng separate from the game object. Before, you couldn't initialize a `D6` without providing an external rng.

mrbermell:
> EDIT: Ah, I was not passing the game.rng to the from_string() parsers. Should be good now?

Now there's no direct problem but I still feel that we should remove the possibility to make that mistake. I propose that `self.rng` is removed from the dice classes. 

```python 
class Die(ABC):
   # __init__ removed 

class D6(Die, Immutable):
    def __init__(self, rng): # default argument removed 
        # call to super() removed. 
        if len(D6.FixedRolls) > 0:
            self.value = D6.FixedRolls.pop(0)
        else:
            self.value = rng.randint(1, 7)  # 'self' removed from self.rnd.randint(...) 
```

njustesen:
You are right. There's no reason why this PR should change this. I've reverted it back.

mrbermell:
Awesome, I'll finish my review today.

mrbermell:
Very nice! Two more things (you may delegate this one to me). 

1) in [botbowl/data/config/web-7.json](https://github.com/njustesen/botbowl/blob/feature/small-web/botbowl/data/config/web-7.json) there's a duplication of these lines. 
```
    "name": "botbowl Web App",
    "arena": "ff-pitch-1.txt",
    "ruleset": "BB2016",
    "dungeon": false,
```
Also I don't see a purpose to keep the `web.json` when `web-11.json` is exactly the same. Also I think that the "secondary" time limit in `gym-11.json` should be 10 not 5. 

2) One suggestion. In the Scatter procedure, 
I think it would be cleaner to create a D2 die class. And simply have: 
```python 
distance_roll = DiceRoll.from_string(config.kick_scatter_dice)
distance = distance_roll.get_sum() - int(config.kick_scatter_dice=='d2') 
```

We should then also change the CoinTossFlip procedure to use the D2 die 
instead of sampling directly from the rng as is done now: 

```python 
class CoinTossFlip(Procedure):
    ...
    def step(self, action):
        if action.action_type == ActionType.HEADS:
            if self.game.rng.rand(1)[0] >= 0.5:
                self.game.state.coin_toss_winner = self.game.state.away_team
                self.game.report(Outcome(OutcomeType.HEADS_WON))
 ```

njustesen:
> in [botbowl/data/config/web-7.json](https://github.com/njustesen/botbowl/blob/feature/small-web/botbowl/data/config/web-7.json) there's a duplication of these lines.

Fixed!

> Also I don't see a purpose to keep the web.json when web-11.json is exactly the same.

Fixed!

> Also I think that the "secondary" time limit in gym-11.json should be 10 not 5.

It's actually 20. Should be good now.

> One suggestion. In the Scatter procedure, I think it would be cleaner to create a D2 die class. And simply have:

I actaully did this at first but I don't want to have dice that don't exist. The dealbreaker that made me delete it, was that the web UI have to show an actual die face in the log, and I don't want to invent a 2D die. It requires a bit more code with a D6 though.

njustesen:
Do thumbs up mean it's approved @mrbermell ? :D

mrbermell:
Haha yes it does. I'm on the phone so I can do a proper approve.

PR #224, opened 2022-04-05
Team loading bug fixed

### LINKED ISSUE #223: Gym envs always use the 11-player teams
I am pretty sure we dealt with this previously. Not sure what happened.

It's due to this line in the BotBowlEnv.__init__():
```python
self.home_team = load_team_by_filename('human', self.ruleset, board_size=11)
```

njustesen:
Fixes #223 by using the EnvCong.size property to load the correct team and renamed renamed team files from `data/<board-size>/human-<board-size>.json` to `data/<board-size>/human.json`.

Also changed teams to always have 1 reserve player.

### DIFF
--- a/botbowl/ai/env.py
+++ b/botbowl/ai/env.py
@@ -55,6 +55,7 @@ class EnvConf:
                  extra_feature_layers: Optional[Iterable[FeatureLayer]] = None,
                  pathfinding=False):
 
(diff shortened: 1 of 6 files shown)


mrbermell:
Good that you fixed the teams too. It's an approval from me!

PR #228, opened 2022-04-08
Adds seed to RandomBot in BotBowlEnv

### LINKED ISSUE #70: Random seed is not honored across multiple simulations
# Description

The environment's random seed is not preserved/honored during game setup. In particular, even though I use the same seed in `environment.seed()` I get different results during a seeded randomized setup of the game board. This is unfortunate because it limits reproducibility. 

In other words, running `environment.reset()` and then `environment.seed(12345)` will result in different outcomes after a seeded random setup.

# Possible Fixes

[`FFAIEnv.reset()`](https://github.com/njustesen/ffai/blob/master/ffai/ai/env.py#L364) creates a new random seed. For the purposes of running across different games it is, of course, important that each new game be different. But if someone wanted to repeat the same game then one possible solution is to provide an optional keyword argument:

```python
class FFAIEnv(gym.Env):

    def reset(self, seed=None):
        # ...
        if seed is None:
            seed = self.rnd.randint(0, 2**31)
        game = Game(...
                    seed=seed)
        self.game.init()
        return self._observation(self.game)
```

I haven't yet checked how this seed propagates into the environments random number generators.

# Steps to Reproduce

Below is some code demonstrating how even though a seed is specified for Numpy, the RandomState, and the environment itself multiple executions of this setup script results in different team setups.

```python
import gym
import ffai.ai
import matplotlib.pyplot as plt
import numpy as np

from ffai.ai.env import FFAIEnv

def ffai_random_action_to_setup(environment, seed=None):
    np.random.seed(seed)
    random_state = np.random.RandomState(seed)
    
    # The FFAI environment forgets its seed when reset. Unfortunately, I haven't
    # found a way to fix the seed in such a way that the setup is done the same
    # way each time. Interestingly, the setup is kind of the same across seeds...
    observation = environment.reset()
    environment.seed(seed)
    environment._seed = seed   # sad attempt at making sure the seed is specified
    assert(environment.get_seed() == seed)

    reward = 0
    is_terminal = 0
    in_setup = lambda procedure: procedure[0:6].sum() > 0
    procedure = observation['procedure']

    while in_setup(procedure):
        random_action = get_random_action(environment, random_state)
        observation, reward, is_terminal, _ = environment.step(random_action)
        procedure = observation['procedure']

    return observation, reward, is_terminal, None


def get_random_action(envrironment, random_state):
    action_type = random_state.choice(environment.available_action_types())
    positions = envrironment.available_positions(action_type)
    action_position = random_state.choice(positions) if len(positions) > 0 else None

    random_action = {
        'action-type': action_type,
        'x': action_position.x if action_position is not None else None,
        'y': action_position.y if action_position is not None else None,
    }
    return random_action


environment = gym.make('FFAI-v1')
observation, reward, is_terminal, _ = ffai_random_action_to_setup(environment, seed=12345)

# plot
own_players = observation['board']['own players']
opp_players = observation['board']['opp players']
players = own_players - opp_players
plt.matshow(players, cmap='RdBu', vmin=-1, vmax=1)
```

njustesen:
Fixes #70 by passing the seed to the RandomBot in BotBowlEnv.

### DIFF
--- a/botbowl/ai/env.py
+++ b/botbowl/ai/env.py
@@ -382,8 +382,8 @@ class BotBowlEnv(gym.Env):
         self.game = Game(game_id=str(uuid.uuid1()),
                          home_team=deepcopy(self.home_team),
                          away_team=deepcopy(self.away_team),
-                         home_agent=BotBowlEnv._create_agent(self.home_agent),
-                         away_agent=BotBowlEnv._create_agent(self.away_agent),
+                         home_agent=BotBowlEnv._create_agent(self.home_agent, seed),
+                         away_agent=BotBowlEnv._create_agent(self.away_agent, seed),
                          config=self.env_conf.config,
                          ruleset=self.ruleset,
                          seed=seed)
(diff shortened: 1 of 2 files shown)


mrbermell (approved):
Very nice. approval from me!

Issue #226, opened 2022-04-07
Scripted bot doesn't score when away


njustesen:
<img width="1067" alt="Screenshot 2022-04-07 at 16 44 42" src="https://user-images.githubusercontent.com/1433421/162229350-2ffa33fc-831f-4723-9034-b7564790d219.png">

It actually plays well but seem to move the ball carrier to the top and not not the endzone.
<img width="458" alt="Screenshot 2022-04-07 at 16 44 32" src="https://user-images.githubusercontent.com/1433421/162229375-763bf728-6cac-4a21-acad-e99b9d084c1b.png">

njustesen:
Here's a view from PyCharms debugger where it attempts to find a path to the endzone but finds one to the top.

<img width="1015" alt="Screenshot 2022-04-11 at 09 42 17" src="https://user-images.githubusercontent.com/1433421/162692653-e7163776-084d-4a14-be48-736d9535a8e3.png">


<img width="1141" alt="Screenshot 2022-04-11 at 09 43 42" src="https://user-images.githubusercontent.com/1433421/162692777-ef573c4b-0625-4b6d-a70a-fc2909723ebf.png">

mrbermell:
Solved in #229

PR #229, opened 2022-04-11
Failing tests + minor python pathfinder fix

njustesen:
Adds tests to #226 and a minor fix to the python pathfinder that didn't always pick the shortest path to the endzone.


mrbermell:
Thanks for writing the tests. Made bug chasing much faster. Issue is solved and PR can be merged.

PR #230, opened 2022-04-11
MCTS example and tutorial

njustesen:
Adds a vanilla MCTS implementation and a tutorial to explain it.


mrbermell:
I only have comments about the code I wrote. The hash function. When I'm reviewing it again, I don't see why the player state wouldn't check `player.state.up`. 

Also, it confusing for a user why the position hash doesn't use `f"{hash(position):3.0f}". I suggest we replace Square's hash function with this smaller `self.x*17+self.y`

Finally, there should be a remarked it can only be used for teams with normal skills. Since its not checking for hypnotized and taken root etc..

mrbermell:
I think an xml output is reasonable in the example. I already implemented it with ElementTree in my own implementation and I'm happy to add it here if you agree.

njustesen:
I've attempted to address your comments without refactoring the framework too much. My hash game is not super strong but I believe my changes have improved the performance and the code should be clearer now. What do you think?

> I think an xml output is reasonable in the example. I already implemented it with ElementTree in my own implementation and I'm happy to add it here if you agree.

Nice. feel free to replace the XML generation as long as the code is easy to read.

examples/mcts_example.py
```diff
@@ -0,0 +1,396 @@
...
+        return action
+
+    def end_game(self, game):
+        pass
+
+
+# Register the bot to the framework
+botbowl.register_bot('mcts', MCTSBot)
+
+if __name__ == "__main__":
```
mrbermell, line 330: I think we should use 
```python
def main(): 
    ... 
if __name__ == "__main__":
    main() 
```

examples/mcts_example.py
```diff
@@ -0,0 +1,396 @@
...
+            print("--------------------------")
+            print(f"MCTS wins: {mcts_wins}")
+            print(f"Random wins: {random_wins}")
+            print(f"MCTS TDs: {mcts_tds}")
+            print(f"Random TDs: {random_tds}")
+            mcts_wins_env += mcts_wins
+            random_wins_env += random_wins
+            mcts_tds_env += mcts_tds
+            random_tds_env += random_tds
+            break
```
mrbermell, line 391: I think this `break` statement can be removed.

examples/mcts_example.py
```diff
@@ -0,0 +1,396 @@
...
+                continue
+            if len(action_choice.players) > 0:
+                for player in action_choice.players:
+                    actions.append(Action(action_choice.action_type, position=None, player=player))
+            elif len(action_choice.positions) > 0:
+                for position in action_choice.positions:
+                    actions.append(Action(action_choice.action_type, position=position))
+            else:
+                actions.append(Action(action_choice.action_type))
+        return actions
```
mrbermell, line 128: Could consider shuffling the action list as simple a remedy to home playing better than away.  
```python 
random.shuffle(actions)
return actions
```

examples/mcts_example.py
```diff
@@ -0,0 +1,396 @@
...
+    def print(self, tabs=0):
+        t = ''.join(['\t' for _ in range(tabs)])
+        min_max = 'min' if self.parent.opp() else 'max'
+        if len(self.outcomes) == 0:
+            print(f"{t}<ChanceNode p='{min_max}' visits={self.num_visits()} score={self.score()} action='{self.action.to_json()}'/>")
+        else:
+            print(f"{t}<ChanceNode p='{min_max}' visits={self.num_visits()} score={self.score()} action='{self.action.to_json()}'>")
+            for _, child in self.outcomes.items():
+                child.print(tabs+1)
+        if len(self.outcomes) > 0:
```
mrbermell, line 178:
```suggestion

```
You don't need this `if` check

examples/mcts_example.py
```diff
@@ -0,0 +1,396 @@
...
+        self.action = action
+        self.outcomes = {}  # hash: Node
+        self.evaluations = []
+        self.terminal = False
+
+    def print(self, tabs=0):
+        t = ''.join(['\t' for _ in range(tabs)])
+        min_max = 'min' if self.parent.opp() else 'max'
+        if len(self.outcomes) == 0:
+            print(f"{t}<ChanceNode p='{min_max}' visits={self.num_visits()} score={self.score()} action='{self.action.to_json()}'/>")
```
mrbermell, line 173:
```suggestion
            print(f"{t}<ChanceNode p='{min_max}' visits={self.num_visits()} score={self.score()} action='{self.action}'/>")
```
More pythonic to use the defined __repr__ function  =)

examples/mcts_example.py
```diff
@@ -0,0 +1,396 @@
...
+        else:
+            print(f"{t}<ChanceNode p='{min_max}' visits={self.num_visits()} score={self.score()} action='{self.action.to_json()}'>")
```
mrbermell, line 175:
```suggestion
            print(f"{t}<ChanceNode p='{min_max}' visits={self.num_visits()} score={self.score()} action='{self.action}'>")
```
More pythonic to use the defined repr function =)

mrbermell (changes requested):
Again, wow! This is awesome! I've added smaller comments here and there. My bigger comment is the xml-tree. It's not valid xml I think because not all attributes are wrapped with quote marks. I think it's missed opportunity and a quick fix :relaxed: 

EDIT:  Also, remember to the mcts.md to the tutorials.md so the users can find it

examples/mcts_example.py
njustesen: Done

examples/mcts_example.py
njustesen: Done

examples/mcts_example.py
njustesen: I agree this is a decent approach but it would likely reduce the performance when playing as home. I like keeping these results because it highlights that a strong MCTS implementation needs proper action policy or ordering.

examples/mcts_example.py
njustesen: This is to only make the closing tag if it wasn't closed on line 173, so I think it should be kept.

examples/mcts_example.py
njustesen: Done

PR #234, opened 2022-05-17
bugfix get_safest_path() for handoff actions

mrbermell:
I stumbled over a bug in `get_safest_path()`, it didn't detect that a handoff action was started and didn't return a handoff path as it should. This PR fixes that and includes a test.

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install .
+        pip install . 
     - name: Lint with flake8
       run: |
         pip install flake8


mrbermell:
@njustesen , can you take a look please :)  It's annoying that all github workflow tests fail because one of the flags in the current pip command is deprecated.

njustesen:
I will take a look at the get safest path change soon. Until then, can you make a new PR with the pip fix and merge it?

botbowl/core/pathfinding/python_pathfinding.py
```diff
@@ -557,7 +557,8 @@ def get_safest_path(game, player, position, from_position=None, allow_team_rerol
     """
     if from_position is not None and num_moves_used != 0:
         orig_player, orig_ball = _alter_state(game, player, from_position, num_moves_used)
-    can_handoff = game.is_handoff_available() and game.get_ball_carrier() == player
+    can_handoff = (game.is_handoff_available() or game.get_player_action_type() is PlayerActionType.HANDOFF) and \
+                  game.get_ball_carrier() == player
```
njustesen, line 561: I think this makes sense. My only suggestion is to make this expression a function with an understandable name. 

image

njustesen (changes requested):
Great. You can merge this PR after addressing my small comment.

botbowl/core/pathfinding/python_pathfinding.py
mrbermell: You know how bad I am at coming up with understandable names :laughing:  What should it be?

PR #236, opened 2022-05-18
Improved game._is_action_allowed()

mrbermell:
In game.py, the method `_is_action_allowed()` is renamed to `is_action_allowed()` to make it _public_. 

I've also removed the print statements, which is annoying and confusing. Instead the function returns an object that can be evaluated as bool but also contain an error message which is useful when debugging for example, example usage: 

```python 
if not game.is_action_allowed():  # can be evaluated as bool 
    print(f"{game.is_action_allowed()}")  # can be evaluated as string, will print error message with helpful detail
```

Let me know what you think!

### DIFF
--- a/botbowl/ai/proc_bot.py
+++ b/botbowl/ai/proc_bot.py
@@ -97,10 +97,10 @@ class ProcBot(Agent):
             assert action is not None
 
         # handle illegal action. if handle_illegal_action() returns a new legal action, return that.
-        if not game._is_action_allowed(action):
+        if not game.is_action_allowed(action):
             new_action = self.handle_illegal_action(game, action)
             assert isinstance(new_action, Action)
-            assert game._is_action_allowed(new_action)
+            assert game.is_action_allowed(new_action)
             return new_action
 
         return action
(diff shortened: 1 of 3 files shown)


njustesen:
I think renaming the function to make it public is a good idea but I'd prefer not to do it now. Let's wait for after Bot Bowl IV.

Not sure I am a fan of the BoolWithMsg. There must be a more standard way to solve this issue. What about warnings or maybe make use of the config.debug flag?

njustesen (changes requested):
I think renaming the function to make it public is a good idea but I'd prefer not to do it now. Let's wait for after Bot Bowl IV.

Not sure I am a fan of the BoolWithMsg. There must be a more standard way to solve this issue. What about warnings or maybe make use of the config.debug flag?

mrbermell:
> I think renaming the function to make it public is a good idea but I'd prefer not to do it now. Let's wait for after Bot Bowl IV.

Yes, let's wait. 

> Not sure I am a fan of the BoolWithMsg. There must be a more standard way to solve this issue. What about warnings or maybe make use of the config.debug flag?

Having thought about it, we should just have two functions. I propose this, what do you think? 

```python
def is_action_legal(action) -> bool: 
   ... 

def get_illegal_action_warning(action) -> str: 
    ... 

```

njustesen:
When and how would you use `get_illegal_action_warning(action) -> str:`? 
Is it better than just keeping the warnings and only raise then if `config.debug = True`

PR #237, opened 2022-05-24
Better xml tree generation in mcts example

mrbermell:


examples/mcts_example.py
```diff
@@ -70,7 +71,7 @@ def simple_heuristic(game: botbowl.Game, agent:botbowl.Agent):
             own += 0.1
         elif game.get_winner() is None:
             opp += 0.1
-    return 0.5 + own - opp
```
njustesen, line 73: The heuristic should return a value between 0 and 1 as if it's a prediction of the zero-sum reward at the end of the game. So I think we should keep the 0.5 here.

examples/mcts_example.py
```diff
@@ -125,6 +126,7 @@ def _extract_actions(self, game):
                     actions.append(Action(action_choice.action_type, position=position))
             else:
                 actions.append(Action(action_choice.action_type))
+        random.shuffle(actions)
```
njustesen, line 129: Can you make this optional and disabled by default? I want the script to follow the tutorial.

njustesen (changes requested):
Everything regarding the XML looks good but I have a few comments to the rest.

examples/mcts_example.py
mrbermell: Good point, thanks! It's updated!

PR #250, opened 2022-08-17
Replay button fix

njustesen:
The replay buttons wasn't clickable.
The crowd square bar was overlapping the replay button bar. I made a conditional to move it down if it's replaying.


mrbermell (approved):
Works nice now! :+1:

PR #253, opened 2023-03-28
Preparation for Bot Bowl V

njustesen:
This will update the website with info for Bot Bowl V.
It's basically the same as last year. Anything we want to change?

I also had to remove python 3.6 from our github workflow since it's no longer available.

TODO:
- [x] Check if we still have the $500 prize from modl
- [ ] Make sure we get prize money from IEEE
- [ ] Any news on the Rust forward model we can include?
- [ ] How do we want to run the newsletter / pre-signup thing?


mrbermell (approved):
It looks good, thanks for doing this! 

The Rust engine still needs some a fair amount of work and testing, I'll use some epsilon time to get it up and running. 

I think it's better to announce ASAP and we can add the news letter thing later.

PR #267, opened 2023-11-24
hot seat option now available vs bot.

FatRedBird:
Hot seat should have only been disabled if neither player was human, as opposed to if either player was human.


njustesen:
Hot seat only makes sense if you are two humans playing.

FatRedBird:
Fair enough :) Perhaps I misunderstood something, what was I supposed to do in order to play against a bot?

njustesen:
Can you make a game with one human and one bot? Then click the team name in the overview :)

FatRedBird:
Ah yep, that works thankyou

